PR #9, opened 2025-01-31
[mohammedhamdhan] iP

mohammedhamdhan:


src/main/java/Orca.java
```diff
@@ -0,0 +1,48 @@
+import java.util.Scanner;
+
+public class Orca {
+    public static final String LINE = "----------------------------------------";
```
Lyam-T: Should this be renamed to something more specific?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,73 @@
+public class TaskManager {
+    private final Task[] tasks;
+    private int taskCount;
+
+    public TaskManager(int maxSize) {
```
Lyam-T: Do you meant to make maxSize to be a constant? If so should it be capital instead?

lishunyang12: Constant names must be all uppercase using underscore to separate words. I think it should be MAXSIZE instead.

src/main/java/Orca.java
lishunyang12: I think they should be named as partitionLine instead, which seems to be more descriptive. Just a suggestion

src/main/java/Orca/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+import Orca.tasktype.Todo;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Storage {
+    // Use OS-independent file separators.
```
aditichadha1310, line 20: These comments seem to be for your reference instead of the reader. You may want to erase them

src/main/java/Orca/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+        ArrayList<Task> tasks = new ArrayList<>();
+        File file = new File(DATA_FILE);
+
+        // If the file doesn't exist, return an empty list.
+        if (!file.exists()) {
+            return tasks;
+        }
+
+        try (Scanner scanner = new Scanner(file)) {
+            while (scanner.hasNextLine()) {
```
aditichadha1310, line 59: Avoid the arrow-head style code.

src/main/java/Orca/TaskManager.java
```diff
@@ -0,0 +1,251 @@
...
+            Deadline newEntry = new Deadline(taskDescription);
+            tasks.add(newEntry);
+            saveTasksToStorage();
+            printAddedDeadlineTask(newEntry, deadlineTime);
+        } catch (Exception e) {
+            System.out.println(Orca.LINE + "\nError adding deadline task: " + e.getMessage() + "\n" + Orca.LINE);
+        }
+    }
+
+    public void addEventTask(String taskDescription) {
```
aditichadha1310, line 55: try to keep every method to less than equals to 30 lines of code. Avoid very long methods

PR #10, opened 2025-01-31
[Alaneel] iP

Alaneel:


src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+public class Alan {
+    private static final String DIVIDER = "____________________________________________________________";
+    private static final int MAX_TASKS = 100;
+    private static final String COMMAND_BYE = "bye";
+    private static final String COMMAND_LIST = "list";
+    private static final String COMMAND_MARK = "mark ";
+    private static final String COMMAND_UNMARK = "unmark ";
+    private static final String COMMAND_TODO = "todo ";
+    private static final String COMMAND_DEADLINE = "deadline ";
+    private static final String COMMAND_EVENT = "event ";
```
Jasonlobo21: It is nice to see you created variables for your commands

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+            handleMarkTask(userInput, false);
+        } else if (userInput.startsWith(COMMAND_TODO)) {
+            handleTodo(userInput);
+        } else if (userInput.startsWith(COMMAND_DEADLINE)) {
+            handleDeadline(userInput);
+        } else if (userInput.startsWith(COMMAND_EVENT)) {
+            handleEvent(userInput);
+        } else {
+            printAvailableCommands();
+        }
```
Jasonlobo21: It is nice to see you created methods so as to not clutter your if-else statements

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+        } else if (userInput.startsWith(COMMAND_UNMARK)) {
+            handleMarkTask(userInput, false);
+        } else if (userInput.startsWith(COMMAND_TODO)) {
+            handleTodo(userInput);
+        } else if (userInput.startsWith(COMMAND_DEADLINE)) {
+            handleDeadline(userInput);
+        } else if (userInput.startsWith(COMMAND_EVENT)) {
+            handleEvent(userInput);
+        } else {
+            printAvailableCommands();
```
Jasonlobo21: Perhaps you could try to use switch cases instead of if-else statements

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    private void printAvailableCommands() {
+        System.out.println("Invalid command. Available commands: todo, deadline, event, list, mark, unmark, bye");
+    }
+
+    public static void main(String[] args) {
+        new Alan().start();
+    }
+}
```
Jasonlobo21: Good use of K&R style brackets

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+                System.out.println("  " + tasks[taskNumber]);
+            } else {
+                System.out.println("Invalid task number. Please try again.");
+            }
+        } catch (NumberFormatException e) {
+            System.out.println("Please provide a valid task number.");
+        }
+    }
+
+    private boolean isValidTaskIndex(int index) {
```
Jasonlobo21: Good use of variable names and camelCase

Jasonlobo21 (commented):
Overall, really liked it, it was easy to read and understand!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
+    protected String by;
```
lakshm1i: Instead of by, consider dueDate, which aligns more closely with conventional naming for deadlines.

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+            handleExit();
+            shouldContinue = false;
+        } else if (userInput.equals(COMMAND_LIST)) {
+            handleList();
+        } else if (userInput.startsWith(COMMAND_MARK)) {
+            handleMarkTask(userInput, true);
+        } else if (userInput.startsWith(COMMAND_UNMARK)) {
+            handleMarkTask(userInput, false);
+        } else if (userInput.startsWith(COMMAND_TODO)) {
+            handleTodo(userInput);
```
lakshm1i: Methods like handleTodo, handleDeadline, and handleEvent are descriptive and follow good naming conventions!

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+        System.out.println();
+
+        return shouldContinue;
+    }
+
+    private void handleExit() {
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+    private void handleList() {
```
lakshm1i: Could consider renaming handleList() to something more descriptive like displayTaskList()!

src/main/java/Alan.java
```diff
@@ -0,0 +1,182 @@
...
+            return;
+        }
+
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println((i + 1) + "." + tasks[i]);
+        }
+    }
+
+    private void handleMarkTask(String userInput, boolean markAsDone) {
```
lakshm1i: Maybe could rename to just markTask()? I think handle prefix can be unnecessary and a bit redundant because it's clear that the purpose of these methods is to mark a task.

lakshm1i (commented):
Great job!

src/main/java/alan/Alan.java
```diff
@@ -0,0 +1,377 @@
+package alan;
+
+import alan.core.*;
```
NCF3535: Imported classes should be listed explicitly, even if you are using the entire class. 
For example:
`import java.util.List;`
This gives an excellent documentation value for the class at hand and makes the class easier to comprehend and maintain.

src/main/java/alan/exception/InvalidTimeFormatException.java
```diff
@@ -0,0 +1,7 @@
+package alan.exception;
+
+public class InvalidTimeFormatException extends AlanException {
+    public InvalidTimeFormatException(String type, String format) {
+        super(String.format("Invalid time format for. Expected format: %s", type, format));
```
NCF3535, line 13: There are 2 arguments `type` and `format` here when there is only one `%s` placeholder.  If an exception occurs may crash the program.

src/main/java/alan/Alan.java
```diff
@@ -0,0 +1,377 @@
...
+            throw new IllegalArgumentException("Invalid task format");
+        }
+
+        String type = parts[0];
+        boolean isDone = "1".equals(parts[1]);
+        String description = parts[2];
+
+        Task task;
+        switch (type) {
+            case "T":
```
NCF3535: Note that there should be no indentation for case clauses unless they are lambda-style.

NCF3535 (commented):
Overall the code is quite good and easy to read! Do take note of the minor issues mentioned. Keep it up!

PR #11, opened 2025-01-31
[brand0nnn] iP

brand0nnn:


src/main/java/Dino.java
```diff
@@ -0,0 +1,122 @@
+import java.util.Scanner;
+
+public class Dino {
+    public static void printLine() {
+        System.out.println("____________________________________________________________");
```
H-ZhanHao: Great job creating a function for printLine. Makes the code look a lot more readable

src/main/java/Dino.java
```diff
@@ -0,0 +1,122 @@
...
+        System.out.println("Got it. I've added this task:");
+        System.out.println("  [" + tasks[size].getTypeIcon() + "][ ] " + task);
+        System.out.println("Now you have " + (size + 1) + " tasks in the list.");
+        printLine();
+    }
+
+    public static void main(String[] args) {
+        sayHello();
+
+        Task[] tasks = new Task[100];
```
H-ZhanHao: Magic number. A constant may be a better option here.

src/main/java/Dino.java
```diff
@@ -0,0 +1,122 @@
...
+                size++;
+            } else if (line.startsWith("mark ")) {
+                markItem(tasks, line);
+            } else if (line.startsWith("unmark ")) {
+                unmarkItem(tasks, line);
+            } else {
+                printLine();
+                System.out.println("Im sorry, I don't understand what you mean by that");
+                printLine();
+            }
```
H-ZhanHao: Great job abstracting all the low-level functions. The if-else statements are very readable

src/main/java/Dino.java
```diff
@@ -0,0 +1,122 @@
...
+            System.out.println("OK, I've marked this task as not done yet:");
+            System.out.println("  [ ] " + tasks[val-1].description);
+        }
+        printLine();
+    }
+
+    public static void addItem(Task[] tasks, int size, String line, Type type) {
+        String task = "";
+        String[] parts;
+        switch (type) {
```
H-ZhanHao: Nice idea to use enum to identify types of commands

H-ZhanHao (commented):
The code was very organised and concise. Just some minor nits to fix to make it even better. Great job overall!

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,178 @@
...
+        for (int i = 0; i < size; i++) {
+            String date = tasks[i].getDate();
+            System.out.println(i+1 + ".[" + tasks[i].getTypeIcon() + "][" + tasks[i].getStatusIcon() + "] " + tasks[i].description + date);
+        }
+        printLine();
+    }
+
+    public static void markItem(Task[] tasks, String line, boolean mark) throws DinoException {
+        printLine();
+        int startIndex = mark ? 5 : 7;
```
zavsky: What do the `5` and `7` mean here? Maybe you could consider placing into constant variables for better readability?

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,178 @@
...
+        String number = line.substring(startIndex);
+        int val = Integer.parseInt(number);
```
zavsky: The variable naming does not seem very intuitive as `number` is assigned to a String type while `val` just below is assigned to int type. Maybe consider renaming them?

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,178 @@
...
+                throw new DinoException(ExceptionMessage.EMPTY_COMMAND);
+            }
+            tasks[size] = new Todo(task);
+            break;
+        case DEADLINE:
+            if (line.split(" ", 2)[1].trim().isEmpty()) {
+                throw new DinoException(ExceptionMessage.EMPTY_COMMAND);
+            }
+            parts = line.split("/by", 2);
+            task = parts[0].substring(9).trim();
```
zavsky: Maybe you could consider using a named variable explaining why the `9` is used here? As it seems quite arbitrary and difficult to follow

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,178 @@
...
+    public static void main(String[] args) {
+        sayHello();
+
+        Task[] tasks = new Task[100];
+        int size = 0;
+        boolean isExit = false;
+        String line;
+        Scanner in = new Scanner(System.in);
+
+        while (!isExit) {
```
zavsky, line 31: There is a double negative present in the `while` argument, maybe you can change into a single positive statement to make it easier to understand?

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,178 @@
...
+                } catch (DinoException e) {
+                    System.out.println(e.getMessage());
+                } catch (IndexOutOfBoundsException | NullPointerException e) {
+                    System.out.println("INPUT ERROR: Item number specified does not exist on the list.");
+                }
+            } else {
+                printLine();
+                System.out.println("Im sorry, I don't understand what you mean by that");
+                printLine();
+            }
```
zavsky, line 41: Should the above `if-else` block be written as `switch-case` for better readability?

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,178 @@
...
+        while (!isExit) {
+            line = in.nextLine();
+            if (line.equals("bye")) {
+                isExit = true;
+            } else if (line.equals("list")) {
+                displayList(tasks, size);
+            } else if (line.startsWith("todo ")) {
+                try {
+                    addItem(tasks, size, line, Type.TODO);
+                    size++;
```
zavsky: Should the incrementing of `size` variable be handled within the `addItem()` and related methods?

src/main/java/dino/exception/ExceptionMessage.java
```diff
@@ -0,0 +1,19 @@
+package dino.exception;
+
+public enum ExceptionMessage {
+    EMPTY_COMMAND("INPUT ERROR: dino.Task cannot be empty."),
+    ITEM_MARKED("Oops, this item has already been marked!"),
+    ITEM_UNMARKED("Oops, this item has already been unmarked!"),
+    NO_BY_DATE("INPUT ERROR: No deadline end date specified"),
+    NO_TO_DATE("INPUT ERROR: No event end date specified");
```
zavsky: I like how you have handled the edge cases to better debug or give users an idea of what went wrong in the input

zavsky (commented):
Good job but just a few nits about the magic numbers present throughout. You have used an interesting approach to error detection and handling which should help you in further debugging in the future

src/main/java/dino/Dino.java
```diff
@@ -0,0 +1,46 @@
...
+            } catch (DinoException e) {
+                System.out.println(e.getMessage());
+            }
+        }
+        storage.save(filePath, tasks);
+        ui.sayBye();
+    }
+
+    public static void main(String[] args) {
+        String filePath = "data/savefile.txt";
```
yijiano: Small nit but please follow the convention for string literals [here](https://nus-cs2113-ay2425s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-magic-numbers).

src/main/java/dino/util/Storage.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            Scanner s = new Scanner(f);
+            while (s.hasNext()) {
+                String line = s.nextLine().trim();
+                String[] parts = line.split("\\|", 5);
+                switch (parts[0]) {
+                case "T" -> tasks.add(new Todo(parts[2]));
+                case "D" -> tasks.add(new Deadline(parts[2], LocalDate.parse(parts[4])));
+                case "E" -> tasks.add(new Event(parts[2], LocalDateTime.parse(parts[3]), LocalDateTime.parse(parts[4])));
+                }
```
yijiano: This switch statement might be considered a bit more complicated of an expression, maybe you could split it up and follow the switch case convention as mentioned [here](https://se-education.org/guides/conventions/java/basic.html#layout)

yijiano: Also, perhaps you can consider adding a default case to guard against erroneous input in the storage file.

src/main/java/dino/util/TaskList.java
```diff
@@ -0,0 +1,139 @@
...
+        switch (type) {
+        case TODO:
+            task = line.split(" ", 2)[1].trim();
+            if (task.isEmpty()) {
+                throw new DinoException(ExceptionMessage.EMPTY_COMMAND);
+            }
+            tasks.add(new Todo(task));
+            break;
+        case DEADLINE:
+            if (line.split(" ", 2)[1].trim().isEmpty()) {
```
yijiano: It appears that your checks for valid commands is being applied to all your cases. Perhaps you can consider refactoring this method so that you can reduce the amount of redundant code?

yijiano (commented):
Good work

PR #13, opened 2025-02-01
[Phua Lock Hian] iP

Phua-Lock-Hian:


src/main/java/Echo.java
```diff
@@ -0,0 +1,18 @@
+import java.util.Scanner;
```
samstt: this appears to be a simple echo function, is creating an echo class really necessary?

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task {
```
samstt: i like this, it is very clear and concise

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
+import java.util.Scanner;
+
+public class List {
```
samstt: the logical flow of this code was very clear! good job!

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
...
+            return true;
+        }
+        errorMessage();
+        return false;
+    }
+
+    private void errorMessage() {
+        System.out.println("uh oh, something went wrong meow. try again~");
+        System.out.println("--");
+    }
```
samstt: good job catching the errors in your chatbot

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public void start() {
+        String line;
+        Scanner in = new Scanner(System.in);
+        do {
+            System.out.println("--");
+            line = in.nextLine().trim();
```
samstt: since java is case sensitive, what if the user inputs 'Bye' or 'List'? will your chatbot still be able to run?

src/main/java/Swing.java
```diff
@@ -0,0 +1,12 @@
+//import java.util.Scanner;
+
+public class Swing {
+    public static void main(String[] args) {
```
samstt: main function is very concise. good!

samstt (commented):
Overall, i really like the way your code is presented and i didn't really notice any violations in coding standards. Good job!

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
+import java.util.Scanner;
+
+public class List {
+    private final Task[] list;
+    private int len;
+
+    public List() {
+        list = new Task[100];
```
vincesum: As it is, 100 is a magic number. Have you considered changing the 100 to a constant?

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
...
+    protected String start;
+    protected String end;
+
+    public Event(String description, String start, String end) {
+        super(description);
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
```
vincesum: Nice override annotation, very clear

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
...
+    public void start() {
+        String line;
+        Scanner in = new Scanner(System.in);
+        do {
+            System.out.println("--");
+            line = in.nextLine().trim();
+            String[] parts = line.split(" ", 2);
+            String command = parts[0];
+            System.out.println("--");
+            switch (command) {
```
vincesum: It might be better to implement SLAP by refactoring the code below each case into different methods.

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
...
+            case "unmark":
+                if (!isValidCommand(parts, 2)) break;
+
+                int toUnmark = Integer.parseInt(parts[1]);
+                list[toUnmark - 1].setStatusIcon(false);
+                System.out.println("Aw man, I guess you're not done yet. Do it soon meow:");
+                System.out.println(list[toUnmark - 1].toString());
+                break;
+
+            case "deadline":
```
vincesum: You might want to test whether deadline entered was valid by checking for /by in the second part

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
...
+                String by = deadlineArgParts[1].split(" ", 2)[1];
+                list[len] = new Deadline(deadlineDesc, by);
+                len++;
+                addedToList();
+                break;
+
+            case "event":
+                if (!isValidCommand(parts, 2)) break;
+
+                String[] eventArgParts = parts[1].split("/", 3);
```
vincesum: Have you considered what happens when the person does not follow the format of entering /from and /to and just entered // instead?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
```
vincesum: Good usage of inheritance to reduce code needed.

src/main/java/List.java
```diff
@@ -0,0 +1,109 @@
...
+        list = new Task[100];
+        len = 0;
+    }
+
+    private void addedToList() {
+        System.out.println("added: " + list[len - 1].toString()); //to offset zero indexing
+        System.out.println("Now you have " + len + " tasks in list.");
+    }
+
+    private boolean isValidCommand(String[] parts, int expectedLen) {
```
vincesum: Well done on checking whether command entered is valid!

vincesum (commented):
Overall, well done on your chatbot Swing! There are some conventions that can be better followed and a few bugs but it generally should work well!

src/main/java/Echo.java
Phua-Lock-Hian: it was kept from one of the earlier increments for the iP, I wasn't sure if we were to keep it for grading purposes. but yes, you are right, functionally it is now not needed. thanks!

src/main/java/List.java
Phua-Lock-Hian: good catch! i'll fix that.

src/main/java/List.java
Phua-Lock-Hian: good catch! i'll fix that.

src/main/java/List.java
Phua-Lock-Hian: good catch, that's a bug. i'll fix it.

src/main/java/List.java
Phua-Lock-Hian: agreed, was planning to refactor. thanks!

src/main/java/List.java
Phua-Lock-Hian: good catch, thanks

src/main/java/swing/Parser.java
```diff
@@ -0,0 +1,204 @@
+package swing;
+
+import swing.tasktypes.*;
```
wenxin-c: Please import libraries/classes explicitly instead of using wildcard.

src/main/java/swing/Parser.java
```diff
@@ -0,0 +1,204 @@
...
+        if (index <= 0 || index > tasks.size()) {
+            errorMessage(ui);
+            return true;
+        }
+        return false;
+    }
+
+    private void errorMessage(Ui ui) {
+        ui.showMessage("uh oh, something went wrong meow. try again~");
+    }
```
wenxin-c: You could consider naming the method sound more verb, e.g. ```printErrorMessage```.

src/main/java/swing/Parser.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private void addedToList(TaskList tasks, Ui ui) {
+        ui.showMessage("added: " + tasks.get(tasks.size() - 1).toString()); //to offset zero indexing
+        ui.showMessage("Now you have " + tasks.size() + " tasks in list.");
+    }
```
wenxin-c: Similarly here, a more verb name would be better, e.g. ```addToList```.

src/main/java/swing/Storage.java
```diff
@@ -0,0 +1,114 @@
+package swing;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import swing.tasktypes.*;
```
wenxin-c: Refer to the previous comment on wildcard.

src/main/java/swing/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class Storage {
+
+    /**
+     * Create file for data storage if it does not already exist
+     */
+    public static void createFile(){
+        try {
+            // Create file to store data if it does not exist
+            File f = new File("./data.txt");
```
wenxin-c: You might consider abstracting the file path into a named constant, so that since it's used multiple times, it will be easier for future changes.

src/main/java/swing/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+
+            // Define regex patterns to handle descriptions with spaces:
+            Pattern todoPattern = Pattern.compile("^\\[T\\]\\[( |X)\\] (.*)$");
+            Pattern deadlinePattern = Pattern.compile("^\\[D\\]\\[( |X)\\] (.*?) \\(by: (.*?)\\)$");
+            Pattern eventPattern = Pattern.compile("^\\[E\\]\\[( |X)\\] (.*?) \\(from: (.*?) to: (.*?)\\)$");
+
+            while (scan.hasNextLine()) {
+                String line = scan.nextLine();
+                Matcher matcher;
+                if (line.startsWith("[T]")) {
```
wenxin-c: You might consider extracting the command type and using ```switch``` to check for commands instead of calling ```startsWith``` multiple times.

src/main/java/swing/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+                } else if (line.startsWith("[E]")) {
+                    matcher = eventPattern.matcher(line);
+                    if (matcher.matches()) {
+                        boolean isDone = matcher.group(1).trim().equals("X");
+                        String desc = matcher.group(2);
+                        String from = matcher.group(3);
+                        String to = matcher.group(4);
+                        Event newEvent = new Event(desc, from, to);
+                        newEvent.setStatusIcon(isDone);
+                        taskList.add(newEvent);
```
wenxin-c, line 111: You might consider abstracting each block into a separate method so that code will be neater and more organised.

wenxin-c (commented):
Good job in your IP. There are some comments for your consideration. You could consider abstracting more to make code neater and more organised.

PR #14, opened 2025-02-01
[K-J-Q] iP

K-J-Q:


src/main/java/Deadline.java
```diff
@@ -1,15 +1,17 @@
-public class Deadline extends Task{
+public class Deadline extends Task {
```
flyingapricot: The variable doneBy is declared with default access (package-private), which is not recommended unless explicitly required.

It should be explicitly declared as private

src/main/java/Deadline.java
```diff
@@ -1,15 +1,17 @@
...
     String doneBy = "";
 
-    Deadline(String userInput){
+
+    // Expect user input with one '/'  as seperator for by
+    Deadline(String userInput) {
         userInput = userInput.substring("deadline".length());
```
flyingapricot: Hardcoding "deadline" is not ideal as it may cause issues if the keyword changes.
 
It would be better to define a constant and use it instead.

src/main/java/Event.java
```diff
@@ -1,37 +1,38 @@
...
-    String startDateTime= "";
-    String endDateTime= "";
+public class Event extends Task {
+    String fromDateTime = "";
+    String toDateTime = "";
 
-    Event(String userInput){
+    // Expect user input with three '/'  as separator for from and to
+    Event(String userInput) {
         userInput = userInput.substring("event".length());
```
flyingapricot: Hardcoding "event" is not ideal as it may cause issues if the keyword changes.

It would be better to define a constant and use it instead.

src/main/java/Event.java
```diff
@@ -1,37 +1,38 @@
...
         }
         this.title = userInputs[0];
 
-        for(int i = 1; i < 3; i++){
-            if(userInputs[i].startsWith("from")) {
-                this.startDateTime = userInputs[i].substring(4);
-            } else if (userInputs[i].startsWith("to")){
-                this.endDateTime = userInputs[i].substring(2);
+        for (int i = 1; i < 3; i++) {
+            if (userInputs[i].startsWith("from")) {
```
flyingapricot: It would be better use `.trim()` before checking `startsWith()` to avoid leading spaces affecting the logic.

src/main/java/Minion.java
```diff
@@ -27,7 +27,7 @@ public static void main(String[] args) {
         Task newTask;
         String messageOut;
 
-
+        // Keep reading the user's input until exit condition reached
         while (!toExit) {
             userInput = in.nextLine();
             switch (userInput) {
```
flyingapricot: It was great that you added a default case in the switch statement. This way, you can handle unexpected inputs and improve robustness.

src/main/java/Minion.java
```diff
@@ -5,9 +5,9 @@ public class Minion {
     public static void main(String[] args) {
         // Constant variables
         String logo = "  __  __ _       _             \n"
```
flyingapricot: Constant variables should be declared as `static final` and named using uppercase letters with underscores

flyingapricot (commented):
Thank you for your effort in implementing this! Overall, the code follows good structuring and readability practices. I have suggested some minor improvements related to coding standards, including the use of explicit access modifiers and ensuring consistent formatting. Addressing these will improve maintainability and robustness.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,32 @@
+public class Deadline extends Task {
+    String doneBy = "";
```
yaxuen05: The variable name "doneBy" can a little confusing. Perhaps you could use "duedate" instead?

src/main/java/MessagePrinter.java
```diff
@@ -0,0 +1,26 @@
+public class MessagePrinter {
```
yaxuen05: It was great that you created separate classes. It makes the code much more organised and readable.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,54 @@
+public class TaskList {
+    private Task[] tasks = new Task[100]; // maximum of 100 tasks
```
yaxuen05: 100 seems like a magic number. Perhaps you could declare it as a constant?

src/main/java/TaskList.java
```diff
@@ -0,0 +1,54 @@
...
+    public String markDone(int index) {
+        if (!isValidIndex(index)) {
+            return "Invalid Selection!";
+        }
+        this.tasks[index].setDone(true);
+        return "Nice! I've marked this task as done:\n  " + getTaskString(index);
+    }
+
+    public String unmarkDone(int index) {
+        if (!isValidIndex(index)) {
```
yaxuen05: Maybe you can also consider checking if the user is asking to mark a task that is already marked as done? Likewise for unmarking a task that is already marked as not done.

src/main/java/Minion.java
```diff
@@ -0,0 +1,63 @@
...
+            userInput = in.nextLine();
+            switch (userInput) {
+            case "bye":
+                toExit = true;
+                break;
+            case "list":
+                minionOut.printMessageAndSep(tasks.listTasks());
+                break;
+            default:
+                if (userInput.startsWith("todo")) {
```
yaxuen05: Your code is very structured and neat. It helps a lot with readability.

src/main/java/Event.java
K-J-Q: `.trim()` is used for removing leading spaces after a command, parsing user inputs only.

It is preferable for commands to have a fixed structure (i.e., / to is not allowed).

src/main/java/minion/Minion.java
```diff
@@ -0,0 +1,58 @@
+package minion;
+
+import minion.parser.Parser;
+import minion.parser.UserCommand;
+import minion.task.*;
```
Paulifyer: This does not follow the coding standards given please avoid this import type

src/main/java/minion/Minion.java
```diff
@@ -0,0 +1,58 @@
...
+    public static void main(String[] args) {
+        // Constant variables
+        final String LOGO = """
+                  __  __ _       _
+                 |  \\/  (_)_ __ (_) ___  _ __
+                 | |\\/| | | '_ \\| |/ _ \\| '_ \\
+                 | |  | | | | | | | (_) | | | |
+                 |_|  |_|_|_| |_|_|\\___/|_| |_|
+                """;
+        final String NAME = "Minion";
```
Paulifyer: I think it would make more sense to put these 2 constants into MessagePrinter

src/main/java/minion/Minion.java
```diff
@@ -0,0 +1,58 @@
...
+            minionOut.printMessageAndSep("Unable to load task from memory!");
+        }
+
+
+        // Display logo and messages
+        minionOut.printMessageAndSep("Hello from\n" + LOGO);
+        minionOut.printMessage("Hello! I'm [" + NAME + "]");
+        minionOut.printMessageAndSep("What can I do for you?");
+
+        Boolean toExit = false;
```
Paulifyer: this does not follow the naming convention for booleans

src/main/java/minion/Storage.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+public class Storage {
+    File f = new File(FILE_PATH);
+    private static final String FILE_PATH = "./taskInfo.txt";
```
Paulifyer: The ordering of these 2 variables is confusing

src/main/java/minion/parser/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+                return (Command.DELETE_TASK);
+            } else if (cmd.startsWith("todo")) {
+                return Command.ADD_TODO;
+            } else if (cmd.startsWith("deadline")) {
+                return Command.ADD_DEADLINE;
+            } else if (cmd.startsWith("event")) {
+                return Command.ADD_EVENT;
+            } else {
+                return Command.UNKNOWN;
+            }
```
Paulifyer: why not use an additional method here to not need to nest the logic of the code here

src/main/java/minion/parser/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+        case DELETE_TASK:
+            int deleteIndex = Integer.parseInt(cmd.message.substring("delete".length()).trim()) - 1;
+            minionOut.printMessage(tasks.delete(deleteIndex));
+            return false;
+        case UNKNOWN:
+            minionOut.printMessageAndSep("OOPS!!! I'm sorry, but I don't know what that means :-(");
+            return false;
+        default:
+            throw new MinionException("Keyword not caught!");
+        }
```
Paulifyer: Maybe use seperate methods for each commands execution to shorten the method

src/main/java/minion/task/Deadline.java
```diff
@@ -0,0 +1,53 @@
...
+            String[] userInputs = input.split("/");
+            if (userInputs.length != 2 || userInputs[0].trim().isEmpty()) {
+                throw new MinionException("Ahhh!!! The deadlines must include (1) description and (2) done by datetime");
+            }
+            this.title = userInputs[0].trim();
+            if (userInputs[1].startsWith(KEYWORD_BY)) {
+                this.dueDate = userInputs[1].substring(KEYWORD_BY.length()).trim();
+                if (this.dueDate.isEmpty()) {
+                    throw new MinionException("Ahhh!!! Deadline is empty");
+                }
```
Paulifyer: Perhaps settle the parsing with a method in the parser class?

PR #17, opened 2025-02-01
[LeThiHongMinh] iP

LeThiHongMinh:


src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
+public class Task {
```
denzelloke: i like how you separated each task into its own class, then made an array of classes.

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void markAsNotDone() {
+        isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
```
denzelloke: i like your use of conditionals, makes the code more succinct

src/main/java/Neura.java
```diff
@@ -23,10 +27,27 @@ public static void main(String[] args) {
...
+            } else {
+                // Add the new task to the list
+                if (taskCount < tasks.length) {
+                    tasks[taskCount] = userInput;  // Store the task
+                    taskCount++;  // Increment the task count
+                    System.out.println("____________________________________________________________");
+                    System.out.println("added: " + userInput);
+                    System.out.println("____________________________________________________________");
+                } else {
+                    System.out.println("____________________________________________________________");
```
denzelloke: i like your consideration of this edge cases

src/main/java/Neura.java
```diff
@@ -23,10 +27,27 @@ public static void main(String[] args) {
...
 
-            // Echo the user input
-            System.out.println("____________________________________________________________");
-            System.out.println(userInput);
-            System.out.println("____________________________________________________________");
+            // If the user types "list", display the stored tasks
+            if (userInput.equals("list")) {
+                System.out.println("____________________________________________________________");
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println((i + 1) + ". " + tasks[i]);
```
denzelloke: apparently this is a magic number.
good practice -> 
const int indexOffset = 1;
System.out.println((i + indexOffset) + ". " + tasks[i]);

notsogood practice ->
System.out.println((i + indexOffset) + ". " + tasks[i]);

(didnt know about this either until today)

src/main/java/Neura.java
```diff
@@ -0,0 +1,79 @@
+import java.util.Scanner;
+
+public class Neura {
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+
+        // Store tasks
+        Task[] tasks = new Task[100];  // Array to store up to 100 tasks
```
K-J-Q: It might be good to create a constant to indicate NUM_TASKS = 100

src/main/java/Neura.java
```diff
@@ -0,0 +1,79 @@
...
+        // Loop to keep listening for user input
+        while (true) {
+            String userInput = scanner.nextLine(); // Get user input
+
+            // Exit the loop if the user types "bye"
+            if (userInput.equals("bye")) {
+                System.out.println("____________________________________________________________");
+                System.out.println("Bye. Hope to see you again soon!");
+                System.out.println("____________________________________________________________");
+                break;
```
K-J-Q: I like how break is the first statement. Makes it cleaner!

src/main/java/Neura.java
```diff
@@ -0,0 +1,79 @@
...
+                    tasks[taskIndex].markAsNotDone();
+                    System.out.println("____________________________________________________________");
+                    System.out.println("OK, I've marked this task as not done yet:");
+                    System.out.println("  [" + tasks[taskIndex].getStatusIcon() + "] " + tasks[taskIndex].getDescription());
+                    System.out.println("____________________________________________________________");
+                }
+            }
+            // Add a new task
+            else {
+                if (taskCount < tasks.length) {
```
K-J-Q: What if the user enters something different by mistake? Will it add it as a new task?

src/main/java/Task.java
K-J-Q: The use of conditionals is effective.

src/main/java/Neura.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+
+        // Store tasks
+        Task[] tasks = new Task[100];  // Array to store up to 100 tasks
+        int taskCount = 0;  // Keep track of how many tasks are stored
+
+        // Display initial greeting
+        System.out.println("____________________________________________________________");
```
K-J-Q: This can be made to be a constant, otherwise other maintainers will have to copy this string (or count the number of "_")

K-J-Q (commented):
Overall, great pull request. Just some refactoring needed

src/main/java/neura/exception/NeuraException.java
```diff
@@ -0,0 +1,7 @@
+package neura.exception;
+
+public class NeuraException extends Exception {
+    public NeuraException(String message) {
+        super(message);
+    }
+}
```
okkhoy: prefer to add a line at the end of the file (git/github won't crib then!)

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
+package neura.ui;
+
+import neura.exception.NeuraException;
+import neura.task.Deadline;
+import neura.task.Event;
+import neura.task.Task;
+import neura.task.Todo;
+
+import java.io.*;
```
okkhoy: do not use wildcard imports. import only those class that are necessary

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+public class Neura {
+
+    private static final String EXIT_COMMAND = "bye";
+    private static final String LIST_COMMAND = "list";
+    private static final String MARK_COMMAND = "mark";
+    private static final String UNMARK_COMMAND = "unmark";
+    private static final String TODO_COMMAND = "todo";
+    private static final String DEADLINE_COMMAND = "deadline";
+    private static final String EVENT_COMMAND = "event";
+    private static final String DELETE_COMMAND = "delete";
```
okkhoy: consider naming `COMMAND_XYZ`

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+    private static final String TASK_LIST_FULL_MESSAGE = "Task list is full. Cannot add more tasks.";
+    private static final String SEPARATOR = "____________________________________________________________";  // Constant for separator line
+    private static final String FILE_PATH = "./data/neura.txt";  // Path to save/load tasks
+
+    private ArrayList<Task> tasks = new ArrayList<>();  // Using ArrayList instead of array
```
okkhoy: all these lines have redundant comments

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+    private void displayTasks() {
+        System.out.println(SEPARATOR);
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < tasks.size(); i++) {
+            System.out.println((i + 1) + "." + tasks.get(i));
+        }
+        System.out.println(SEPARATOR);
+    }
+
+    private void markTaskAsDone(String userInput) throws NeuraException {
```
okkhoy: all non-trivial methods (and classes) need header comments

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    private void addTodoTask(String userInput) throws NeuraException {
+        // Ensure that the description after the "todo" command is not empty
+        if (userInput.trim().equals(TODO_COMMAND)) {
+            throw new NeuraException("The description of a todo cannot be empty.");
+        }
+
+        // Extract the description after the "todo" command
+        String description = userInput.substring(TODO_COMMAND.length()).trim();  // Skip "todo" and trim extra spaces
```
okkhoy: redundant comment

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+            saveTasks();  // Save tasks to file after adding
+        } catch (IndexOutOfBoundsException e) {
+            throw new NeuraException("Invalid deadline format. Use: deadline [task] /by [time]");
+        }
+    }
+
+    private void addEventTask(String userInput) throws NeuraException {
+        try {
+            String taskDetails = userInput.substring(6).trim();
+            String[] details = taskDetails.split(" /from ", 2);
```
okkhoy: consider extracting low-level string operations to separate methods to improve SLAP

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+            if (details.length < 2) throw new NeuraException("Invalid event format. Use: event [task] /from [start] /to [end]");
```
okkhoy: our coding standards don't allow single line if statements. conditional statements should be enclosed in braces and must span multiple lines.

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+        File dataFile = new File(FILE_PATH);
+        if (dataFile.exists()) {
+            try (BufferedReader reader = new BufferedReader(new FileReader(dataFile))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    Task task = parseTask(line);
+                    if (task != null) {
+                        tasks.add(task);
+                    }
+                }
```
okkhoy: think if there is any way to eliminate the arrow head code

src/main/java/neura/ui/Neura.java
```diff
@@ -0,0 +1,277 @@
...
+        } else if (task instanceof Deadline) {
+            Deadline deadline = (Deadline) task;
+            return "D | " + (task.isDone() ? "1" : "0") + " | " + task.getDescription() + " | " + deadline.getBy();
+        } else if (task instanceof Event) {
+            Event event = (Event) task;
+            return "E | " + (task.isDone() ? "1" : "0") + " | " + task.getDescription() + " | " + event.getFrom() + " | " + event.getTo();
+        }
+        return "";
+    }
+}
```
okkhoy: the main class is doing all the work. look at the increments A-Packages and A-MoreOOP and see if you can extract related methods & data to separate classes.

okkhoy (commented):
@LeThiHongMinh 
the existing code is generally neat. however, you have some coding standard violations; please try to correct them. another suggestion is to extract multiple classes for related data & code so that the main class doesn't have to do all the heavy lifting

src/main/java/neura/ui/Neura.java
```diff
@@ -66,7 +71,7 @@ private void printGoodbye() {
         System.out.println(SEPARATOR);
     }
 
-    private void handleCommand(String userInput) throws NeuraException{
+    private void handleCommand(String userInput) throws NeuraException {
```
denzelloke: this command is quite complex.
maybe refactoring it would help with maintenance and progression?

src/main/java/neura/task/Task.java
```diff
@@ -1,8 +1,8 @@
 package neura.task;
 
 public class Task {
-    private final String description;
-    private boolean isDone;
+    protected final String description;
```
denzelloke, line 8: could consider using getter functions to maintain encapsulation

src/main/java/neura/ui/Neura.java
```diff
@@ -79,6 +84,8 @@ private void handleCommand(String userInput) throws NeuraException{
             addDeadlineTask(userInput);
         } else if (userInput.startsWith(EVENT_COMMAND)) {
             addEventTask(userInput);
+        } else if (userInput.startsWith(DELETE_COMMAND)) {
+            deleteTask(userInput);
         } else {
             throw new NeuraException("I'm sorry, but I don't know what that means :-(");
```
denzelloke: perhaps switch cases here would be more succinct & readable?
(more of a stylistic decision)

switch (getCommand(userInput)) {
    case "list" -> displayTasks();
    case "mark" -> markTaskAsDone(userInput);
    case "unmark" -> unmarkTaskAsDone(userInput);
    case "delete" -> deleteTask(userInput);
    default -> throw new NeuraException("I'm sorry, but I don't know what that means :-(");
}

src/main/java/neura/ui/Neura.java
```diff
@@ -181,14 +179,89 @@ private void addEventTask(String userInput) throws NeuraException{
...
+                return new Todo(parts[2]);
+            case "D":
+                return new Deadline(parts[2], parts[3]);
+            case "E":
+                return new Event(parts[2], parts[3], parts[4]);
+            default:
+                return null;
+            }
+        } catch (ArrayIndexOutOfBoundsException e) {
+            System.out.println("Error parsing line: " + line);
```
denzelloke: could consider throwing exception / logging error here instead of printing and returning null
or even introducing custom exceptions

src/main/java/neura/ui/Neura.java
```diff
@@ -181,14 +179,89 @@ private void addEventTask(String userInput) throws NeuraException{
...
+        } else if (task instanceof Deadline) {
+            Deadline deadline = (Deadline) task;
+            return "D | " + (task.isDone() ? "1" : "0") + " | " + task.getDescription() + " | " + deadline.getBy();
+        } else if (task instanceof Event) {
+            Event event = (Event) task;
+            return "E | " + (task.isDone() ? "1" : "0") + " | " + task.getDescription() + " | " + event.getFrom() + " | " + event.getTo();
+        }
+        return "";
     }
 }
```
denzelloke: could try making the main class more of a coordinator of all the different functions, rather than the one that does all the work.

for eg, you could have a parser(process input) class, a class for printing output, etc etc

src/main/java/neura/ui/Neura.java
LeThiHongMinh: I removed the wildcard imports. Thank you for your suggestion!

PR #18, opened 2025-02-01
[LEESY02] iP

LEESY02:
Lee Seung-Yoon


src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+ * lowerCamelCase for variables and object instances
+ * UpperCamelCase for Class names
+ * ALL_CAPS_WITH_UNDERSCORE for constants
+ * Print messages encapsulated in private static methods for readability
+ */
+
+public class SeungYoon {
+    private static int numberOfTasks = 0;
+
+    public static final String NEW_LINE = "____________________________________________________________\n";
```
Sean2110: i like how you used snake case for the constant values

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+    public static final String NEW_LINE_TAB = "\t____________________________________________________________\n";
+    public static final String LINE_TAB = "\t____________________________________________________________";
+
+    // "list" command: Print
+    private static void printTaskList(Task[] taskList) {
+        System.out.print(NEW_LINE_TAB);
+        for (int i = 0; i < taskList.length; i += 1) {
```
Sean2110: i like how u used camel case

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private static void printAddTaskMessage(Task task) {
+        System.out.println(
+            NEW_LINE_TAB + "\tadded: " + task.toString() + "\n" + NEW_LINE_TAB
+        );
+    }
+    // end of add
+
+    // "mark" / "unmark" command: Mark as done or undone
+    private static void flipCompletionStatus(Task[] taskList, int taskIndex, boolean completionStatus) {
```
Sean2110: for your boolean you could phrase it in a way that uses prefixes like is or has

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+    private static Task[] executeCommand(Task[] taskList, String input) {
+        Scanner sc = new Scanner(input);
+        String command = sc.next();
+        String task = "";
+        if (sc.hasNextLine()) {
+            task = sc.nextLine().strip();
+        }
+        sc.close();
+
+        switch (command) {
```
Sean2110: i like that u used switch instead of many nested loops

Sean2110 (commented):
good job

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private static void markOrUnmark(Task[] taskList, String indexString, boolean completionStatus) {
+        if (!indexString.isBlank()) {
+            int index = Integer.parseInt(indexString) - 1;
+            flipCompletionStatus(taskList, index, completionStatus);
+        }
+    }
+    // end of command execution algorithm
+
+    public static void main(String[] args) {
```
hyunjinkim1112, line 23: You may want to put the main method at the top of the class so that the reader can easily follow your code.

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+        return taskList;
+    }
+
+    private static Task returnCorrectTask(String input, int index) {
+        Scanner sc = new Scanner(input);
+        String taskType = sc.next();
+
+        switch (taskType) {
+        case "todo":
+            return returnToDo(sc, index);
```
hyunjinkim1112: The method name and the argument name is not every explicit. You may want to consider that into more explicit names like from, to, by, description, idx of what, etc.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+    private final String deadline;
+
+    public Deadline(String task, int index, String deadline) {
+        super(task, index);
+        this.deadline = deadline;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString().replaceFirst("\\[", "[D][") + String.format(" (by: %s)", deadline);
```
hyunjinkim1112: To get the most advantage of polymorphism, instead of using replaceFirst and hardcoding the value, we may consider using getType (which will return you D, T, E based on type of your task).

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,194 @@
...
+            markOrUnmark(taskList, task, false);
+            break;
+        default:
+            taskList = addTask(taskList, input);
+        }
+
+        return taskList;
+    }
+
+    private static void markOrUnmark(Task[] taskList, String indexString, boolean completionStatus) {
```
hyunjinkim1112: The functionality of this function is a bit confusing due to the usage of the language and variables name like markOrUnmark and flipCompletionStatus. You may want to consider writing two methods for mark and unmark respectively.

hyunjinkim1112 (commented):
Overall, good code with al the functionalities required. You can refactor a bit to make it a bit more straightforward!

src/main/java/SeungYoon.java
```diff
@@ -0,0 +1,41 @@
+import static data.Constants.*;
+
+/*
+ * Coding standard is defined by IDE
+ * lowerCamelCase for variables and object instances
+ * UpperCamelCase for Class names
+ * ALL_CAPS_WITH_UNDERSCORE for constants
+ * Print messages encapsulated in private static methods for readability
```
aditichadha1310: These comments seem as if they are written for your reference and not for the reader.

aditichadha1310 (commented):
Overall, very good work! keep it up

PR #19, opened 2025-02-01
[Nora Daza] iP

noradazaperez:


src/main/java/Mimi.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public static void main(String[] args) {
+        String userinput;
+        greet();
+        Scanner sc = new Scanner(System.in);
+        do {
+            userinput = sc.nextLine();
+            System.out.println(bar);
+            String [] input = userinput.split(" ",2);
+
```
trunghjieu2506: Is input[1] always present? I am thinking of cases when users input incomplete command such as "mark". It might be better to do input check for these cases.

src/main/java/Mimi.java
```diff
@@ -0,0 +1,121 @@
+import java.util.Scanner;
+
```
trunghjieu2506, line 1: Great code overall. You might consider add Javadoc comments in the code for better readability when the iP code size gets larger.

src/main/java/Mimi.java
```diff
@@ -0,0 +1,121 @@
...
+                    unmark(Integer.parseInt(input[1]));
+                    break;
+                case "todo":
+                    addToDo(input[1]);
+                    break;
+                case "deadline":
+                    addDeadline(input[1]);
+                    break;
+                case "event":
+                    addEvent(input[1]);
```
trunghjieu2506: There should be no indentation for case clauses according to the coding standard.

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
+import java.util.ArrayList;
+import java.io.*;
```
okkhoy: don't use wildcard imports. import only those classes that are necessary

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+        } catch (MimiException e) {
+            System.out.println(e.getMessage());
+        }
+
+        while (true) {
+            System.out.print("> ");
+            userinput = sc.nextLine();
+            System.out.println(bar);
+
+            // exit if the user types something like 'quit' or 'exit
```
okkhoy: invalid comment. you are only checking for `exit` not `quit`. so the comment is tripping the reader!

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+            if (userinput.equalsIgnoreCase("exit")) {
+                System.out.println("Mimi says adios to you...");
+                break;
+            }
+
+            try {
+                // process the user input
```
okkhoy: redundant comment; the method name next line is self explanatory

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+
+                // Basic validation
+                if (parts.length < 3) {
+                    throw new MimiException("Oh no... There is a PROBLEMA with the file");
+                }
+
+                // Convert "Y"/"N" to boolean
+                boolean done = parts[0].trim().equalsIgnoreCase("Y");
+
+                // The third field is always the description
```
okkhoy: better to identify it by a constant(used as an index) or a specific variable to capture the description, rather than using the index of the array.

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+                String description = parts[2].trim();
+                Task task;
+
+                // The second field is the type (T/D/E)
+                String type = parts[1].trim();
+                if (type.equalsIgnoreCase("T")) {
+                    task = new Task(description, done);
+                }
+                else if (type.equalsIgnoreCase("D")) {
```
okkhoy: else/else if has to be cuddled between the braces.  like: 
`} else { `

pl fix all such occurrences.

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+                else if (type.equalsIgnoreCase("E")) {
+                    // Validation
+                    if (parts.length < 5) {
+                        throw new MimiException("Oh no... There is a PROBLEMA with the file");
+                    }
+                    task = new Event(description, done, parts[3].trim(), parts[4].trim());
+                }
+                else {
+                    // ERROR
+                    throw new MimiException("Oh no... There is a PROBLEMA with the file");
```
okkhoy: all error messages seem to be talking about problem with the file. is that correct? it may be better to have error messages that are specific to the problem encountered

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+        switch (command) {
+            case "list":
+                showList();
+                break;
+            case "mark":
+                // parts[1] must exist, the description
+                if (parts.length == 1) {
+                    throw new MimiException("Oh no... There is a PROBLEMA with the description.");
+                }
+                try{
```
okkhoy: our case blocks should not be indented. 
you can get the IDE to fix this issue easily; check how that is done in one of the early lecture recordings.

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+        if (index > tasks.size()) {
+            throw new MimiException("Sorry! The task doesn't exist. Index out of bounds: " + index);
+        }
+        Task task = tasks.remove(index-1);
+        System.out.println("Ay no! You don't want to do the task anymore? I have removed the next task:");
+        System.out.println(task);
+
+        // Only prints if there are tasks
+        if (!tasks.isEmpty())
+        {
```
okkhoy: coding standard violation here (hint: how should the braces appear in this case?)

src/main/java/Task.java
```diff
@@ -0,0 +1,40 @@
+public class Task {
```
okkhoy: you seem to be missing increments where we ask you to use inheritance for the different task types

src/main/java/Mimi.java
```diff
@@ -0,0 +1,300 @@
...
+        System.out.println("Now you have " + tasks.size() + " tasks");
+    }
+
+    public static void showList() {
+        for (int i = 0; i < tasks.size(); i++) {
+            System.out.println((i+1) + ". " + tasks.get(i).toString());
+        }
+
+    }
+}
```
okkhoy, line 64: the main class is doing too many things. consider extracting out related code and functionality to other classes.
see A-MoreOOP increment

okkhoy (commented):
@noradazaperez 
You have quite a few coding standard violations in your codebase. Try to fix as many as possible before the submission.

PR #22, opened 2025-02-01
[siqiangz] iP

siqiangz:


src/main/java/Pookie.java
```diff
@@ -0,0 +1,38 @@
+import java.util.Scanner;
+
+public class Pookie {
+    private static boolean onOffSwitch = true;
```
Yikbing: naming for boolean is a little ambiguous, could change it to start with: "is", "has", "was"

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,129 @@
+public class TaskManager {
+    private static int listSize = 0;
+    private static final int MAX_LIST_SIZE = 100;
```
Yikbing: good job for replacing the magic literal with symbolic constant!

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,129 @@
...
+        Pookie.doLineBreak();
+    }
+
+    private static void setTaskAsUndone(Task t) {
+        t.setIsDone(false);
+        System.out.println("\tOK, I've marked this task as not done:");
+        System.out.println("\t\t[" + t.getStatusIcon() + "] " + t.getTaskDescription());
+    }
+
+    private static void setTaskAsDone(Task t) {
```
Yikbing: naming could be improved by changing the name to "task" instead of a single letter :D

src/main/java/Pookie.java
siqiangz: got it, thank you for the suggestion!

src/main/java/TaskManager.java
siqiangz: thank you!

src/main/java/TaskManager.java
siqiangz: thank you! will work on it

src/main/java/Parser.java
```diff
@@ -0,0 +1,36 @@
+public class Parser {
+
+    public static void interpretLineInput(String lineInput) {
+        String[] wordArray = separateInput(lineInput);
+
+        switch (wordArray[0]) {
+        case "mark":
```
Yikbing: good job for having no indentations in the case clauses!

siqiangz: thank you!

src/main/java/pookie/files/FileManager.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Scanner;
+
+import pookie.errors.ErrorHandler;
+import pookie.tasks.Task;
+import pookie.tasks.TaskManager;
+
+public class FileManager {
+
+    public static void getSaveFileOnStartup() {
+        File saveFile = new File("./tasks.txt");
```
wenxin-c, line 26: You may consider converting the file path to a named constant so that it will be easier for future changes.

src/main/java/pookie/tasks/MarkTestChecker.java
```diff
@@ -0,0 +1,40 @@
+package pookie.tasks;
+
+import pookie.errors.ErrorHandler;
+import pookie.errors.InvalidMarkTaskException;
+
+public class MarkTestChecker {
```
wenxin-c, line 9: This class is mostly the same as the above ```DeleteTaskChecker``` except some minor details. You may consider combining the checkers instead of duplicating most of the code.

wenxin-c (commented):
Overall well done in your IP and good use of OOP as well. Just minor suggestions for your consideration. Keep it up!

src/main/java/pookie/tasks/TaskManager.java
```diff
@@ -0,0 +1,168 @@
...
+        // For case "event":
+        default:
+            int indexOfFrom = getIndexOfTimeline(wordArray, "from");
+            int indexOfTo = getIndexOfTimeline(wordArray, "to");
+            Event event = new Event(combineString(wordArray, 1, indexOfFrom),
+                    combineString(wordArray, indexOfFrom + 1, indexOfTo),
+                    combineString(wordArray, indexOfTo + 1, wordArray.length));
+            taskList.add(event);
+            return event.getTaskDescription() + " from " + event.getFrom() + " to " + event.getTo();
+        }
```
wenxin-c: You could consider adding event into a case clause and use default to handle unknown commands if there's any.

PR #24, opened 2025-02-01
[jiajun2002] iP

jiajun2002:


src/main/java/Lebron.java
matthewyeo1, line 1: + Good choice of access modifiers for your variables & functions
+ Modularity in splitting your task functions 
- Would be better if you can use macros to define your 'magic' strings in your code
- Create another file to add your task array manipulation functions instead of cluttering the main file

src/main/java/Lebron.java
```diff
@@ -0,0 +1,147 @@
...
+        System.out.println("Be confident in your ability! Current tasks: " + tasksSize);
+        printLinebreak();
+    }
+
+    public static void printErrorMessage(String error){
+        System.out.println("Sorry that's a foul! " + error);
+        printLinebreak();
+    }
+
+    public static void addTodo(String todo) {
```
ManZ9802: Instead of 3 methods accomplishing the same thing, consider using polymorphism of the Task class, (eg. one method - addTask that accomplishes different things based on the subclass)

src/main/java/Lebron.java
```diff
@@ -0,0 +1,147 @@
...
+                break;
+            }
+            if (line.equals("list")) {
+                printTasks();
+                continue;
+            }
+
+            String[] words = line.split(" ");
+            int idx;
+            switch(words[0]) {
```
ManZ9802: good use of switch case for diff commands

src/main/java/Lebron.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        printLinebreak();
+    }
+
+    public static void main(String[] args) {
+        printGreetingMessage();
+        Scanner input = new Scanner(System.in);
+        while(true) {
+            String line = input.nextLine();
+            if (line.equals("bye")) {
```
wongyihao0506: Perhaps it will be good to use ignorecase?

src/main/java/Lebron.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public static void addEvent(String event) {
+        int firstSplit = event.indexOf("/from");
+        int secondSplit = event.indexOf("/to");
+        if ((firstSplit == -1) || (secondSplit == -1)) {
+            printErrorMessage("Your event has to include '/from' and '/to'");
+            return;
+        }
+        int taskIdx = "event ".length();
+        int fromIdx = firstSplit + "/from ".length();
```
wongyihao0506: may be good to rename to make 'fromIdx' and 'toIdx' as the names are not telling of what this variable stores

src/main/java/Lebron.java
wongyihao0506: good switch block, covers all ground

src/main/java/Lebron.java
```diff
@@ -0,0 +1,147 @@
...
+                addEvent(line);
+                break;
+            default:
+                System.out.println("Personally, I think that's a bad play. Try starting with 'todo', 'deadline', or 'event'!");
+                printLinebreak();
+                break;
+            }
+        }
+    }
+}
```
wongyihao0506: overall good and appropriate indentations

src/main/java/lebron/ui/Lebron.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final int MAX_TASK_SIZE = 100;
+    private static Task[] tasks = new Task[MAX_TASK_SIZE];
+    private static int tasksSize = 0;
+
+    public static void printLinebreak() {
+        System.out.println(LINE_BREAK);
+    }
+
+    public static void printGreetingMessage() {
+        System.out.println("Hello! I'm Lebron, just a little kid from Akron");
```
bljhty: Try to avoid the use of magic literals and make them a constant instead

src/main/java/lebron/ui/Lebron.java
```diff
@@ -0,0 +1,202 @@
...
+        System.out.println("What we doing today?");
+        printLinebreak();
+    }
+
+    public static void printAddingMessage(Task[] tasks, int size) {
+        System.out.println("Alright! Get ready to do this: ");
+        System.out.println("--> " + tasks[size - 1]);
+        System.out.println("Be confident in your ability! Current tasks: " + tasksSize);
```
bljhty: Try to avoid the use of magic literals and make them a constant instead

src/main/java/lebron/ui/Lebron.java
```diff
@@ -0,0 +1,202 @@
...
+            break;
+        default:
+            throw new UnknownCommandException();
+        }
+    }
+
+    public static void main(String[] args) {
+        printGreetingMessage();
+        Scanner input = new Scanner(System.in);
+        while(true) {
```
bljhty: you may want to have a more explicit exit condition like a boolean flag instead as while(true) Is generally not advised

src/main/java/lebron/ui/Lebron.java
```diff
@@ -0,0 +1,202 @@
...
+                break;
+            }
+            if (line.equalsIgnoreCase("list")) {
+                printTasks();
+                continue;
+            }
+            try {
+                processCommand(line);
+            } catch (UnknownCommandException e) {
+                printErrorMessage("Not sure what you're trying to make me do, start off with todo/deadline/event instead!");
```
bljhty: magic strings

PR #28, opened 2025-02-01
[Charly2312] iP

Charly2312:


src/main/java/Print.java
```diff
@@ -0,0 +1,80 @@
...
+        System.out.println("-".repeat(length));
+        System.out.println("Bye. Hope to see you again soon!");
+        System.out.println("-".repeat(length));
+    }
+
+    public static void printGreeting() {
+        int length = LINE_DASH_LENGTH; // Length of the horizontal line
+        System.out.println("Hello! I'm Aples");
+        System.out.println("What can I do for you?");
+        System.out.println("-".repeat(length));
```
syCHEN1645: Perhaps can directly call LINE_DASH_LENGTH instead of assigning it to 'int length'? Just like how you did it in the other methods.

src/main/java/Print.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public static void echoTask(Task task, int itemNumber) {
+        System.out.println("-".repeat(LINE_DASH_LENGTH));
+        System.out.println("Got it. I've added this task:");
+        System.out.println("  " + task.toString());
+        if (itemNumber == 1) {
+            System.out.println("Now you have " + itemNumber + " task in the list.");
+        } else {
+            System.out.println("Now you have " + itemNumber + " tasks in the list.");
```
syCHEN1645: I like how you managed single/plural cases!

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
...
+    public static void main(String[] args) {
+        String logo =
+                "                                               \n" +
+                " █████╗ ██████╗ ██╗     ███████╗███████╗       \n" +
+                "██╔══██╗██╔══██╗██║     ██╔════╝██╔════╝       \n" +
+                "███████║██████╔╝██║     █████╗  ███████╗       \n" +
+                "██╔══██║██╔═══╝ ██║     ██╔══╝  ╚════██║       \n" +
+                "██║  ██║██║     ███████╗███████╗███████║       \n" +
+                "╚═╝  ╚═╝╚═╝     ╚══════╝╚══════╝╚══════╝       \n" +
+                "    Aples chatbot at your service              \n";
```
syCHEN1645: Really cool logo!

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
...
+        System.out.println(logo);
+        Task[] list = new Task[MAX_ENTRIES]; //make an array of class type: Task
+        int itemIndex = 0; //index of the item to be added to the list
+        Print.printGreeting();
+
+        while (true) {
```
syCHEN1645, line 32: Maybe extract this while loop to a new method to look more tidy?

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
+import org.w3c.dom.ls.LSInput;
```
rodi-314: May I know what is this imported library for? Not sure where it's used hmm...
```suggestion
import org.w3c.dom.ls.LSInput;
```

src/main/java/HandleTask.java
```diff
@@ -0,0 +1,5 @@
+import java.util.Scanner;
```
rodi-314: Is this class required? It seems empty... ://

src/main/java/Aples.java
rodi-314, line 32: I think a boolean can be created and pass into the while loop instead; while (true) loops are generally not recommended as it may cause infinite loops! :0
```suggestion
        while (true) {
```

src/main/java/Print.java
```diff
@@ -0,0 +1,80 @@
+public class Print {
```
rodi-314: I think it's interesting to create a class just for the print functions!

rodi-314 (commented):
Good job overall! Just some minor tweaks to consider :))

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Arrays;
+import java.util.Scanner;
+
+public class Aples {
+
+    public static final int MAX_ENTRIES = 100;
+    public static final int LINE_DASH_LENGTH = 50;
+
+    public static void handleTodoTask(String line, Task[] list, int itemIndex) {
+        String description = line.substring(5); //take out the description without the category word
```
RyanLing169: magic number 5 not as clearly expressed as other numbers before

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
...
+        list[itemIndex] = new Todo(description);
+        Print.echoTask(list[itemIndex], itemIndex + 1);
+    }
+
+    public static void handleDeadlineTask(String line, Task[] list, int itemIndex) {
+        if (line.contains("/by")) {
+            int byPosition = line.indexOf("/by");
+            String description = line.substring(9, byPosition);
```
RyanLing169: magic number 9 here as well

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
...
+            String deadline = line.substring(byPosition + 4);
```
RyanLing169: magic number 4

src/main/java/Aples.java
```diff
@@ -0,0 +1,98 @@
...
+            list[itemIndex] = new Deadline(description, deadline);
+            Print.echoTask(list[itemIndex], itemIndex + 1);
+        } else {
+            System.out.println("-".repeat(Print.LINE_DASH_LENGTH));
+            System.out.println("Please key in a Deadline task with the keyword '/by'");
+            System.out.println("-".repeat(Print.LINE_DASH_LENGTH));
+            String newLine;
+            Scanner in = new Scanner(System.in);
+            newLine = in.nextLine();
+            handleDeadlineTask(newLine, list, itemIndex);
```
RyanLing169: possibly different levels of abstraction grouped together with line 29 to 31

src/main/java/Aples.java
```diff
@@ -0,0 +1,52 @@
+import Aples.Print.Print;
```
bljhty: Names representing packages should be in all lower case

src/main/java/Aples.java
```diff
@@ -0,0 +1,52 @@
...
+        System.out.println(logo);*/
+        Task[] list = new Task[MAX_ENTRIES]; //make an array of class type: Task
+        int itemIndex = 0; //index of the item to be added to the list
+        Print.printGreeting();
+
+        String line;
+        Scanner in = new Scanner(System.in); //needs to be outside of the loop to ensure the automated test runs
+        while (true) {
+            line = in.nextLine();
+            boolean isBye = line.contains("bye") || line.contains("Bye");
```
bljhty, line 34: String comparisons should be case-insensitive

Consider making magic values like "bye" as a constant?

src/main/java/Aples.java
```diff
@@ -0,0 +1,52 @@
...
+                break;
+            } else if (line.equals("list") || line.equals("List")) {
+                Print.printList(list);
+            } else if (line.contains("unmark")) {
+                Print.unmarkBox(line, list);
+            } else if (line.contains("mark")) {
+                Print.markBox(line, list);
+            } else if (line.contains("todo")) {
+                HandleTask.handleTodoTask(line, list, itemIndex);
+                itemIndex += 1;
```
bljhty: Naming of itemIndex could be a little clearer

src/main/java/Aples.java
bljhty: it would be better to provide a clearer condition as to when to exit the loop instead of just using while(true)

src/main/java/Aples.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static void main(String[] args) {
+        /*String logo =
+                "                                \n" +
+                "    ___    ____  __    ____  ___ \n " +
+                "   /__ | (  _ \(  )  ( ___)/ __) \n " +
+                "  /(__)| |___/ )(__  )__) \__ \ \n " +
+                " (__)(__)|__)  (____)(____)(___/ \n " +
+                "    Aples chatbot at your service   \n";
+        System.out.println(logo);*/
```
bljhty: If this code is meant to be commented out and not used. might want to delete it instead

src/main/java/Aples/Print/Print.java
```diff
@@ -0,0 +1,94 @@
...
+import Aples.Tasks.Task;
+
+public class Print {
+
+    public static final int LINE_DASH_LENGTH = 50;
+
+    public static void printBye() {
+        System.out.println("-".repeat(LINE_DASH_LENGTH));
+        System.out.println("Bye. Hope to see you again soon!");
+        System.out.println("-".repeat(LINE_DASH_LENGTH));
```
bljhty: maybe you might want to consider creating a constant for the '-' dash line instead

src/main/java/Aples/Tasks/HandleTask.java
```diff
@@ -0,0 +1,62 @@
+package Aples.Tasks;
+
+import Aples.Print.Print;
+
+import java.util.Scanner;
+
+public class HandleTask {
```
bljhty: Class names should be a noun

src/main/java/Aples/Tasks/HandleTask.java
```diff
@@ -0,0 +1,62 @@
...
+            Print.echoTask(list[itemIndex], itemIndex + 1);
+        } catch (IndexOutOfBoundsException e) { //take out the description without the category word
+            Print.todoError();
+        }
+    }
+
+    public static void handleDeadlineTask(String line, Task[] list, int itemIndex) {
+        try {
+            //check if there's something after deadline
+            String statement = line.substring(9);
```
bljhty: you may want to try to avoid the use of magic numbers

PR #32, opened 2025-02-02
[Hu Hongheng] iP

Hudou0420:


src/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+                System.out.println("Event " + eventName + " has been unmarked!");
+                System.out.println("--------------------------------------------------");
+                listTasks();
+                System.out.println("--------------------------------------------------");
+                return;
+            }
+        }
+        System.out.println("The event is not in your list!");
+    }
+
```
KeerthanaBanukumar: The class name is good, as it clearly indicates that this class is used for managing a list of tasks (or events, in this case). However, consider renaming eventListSize to taskCount or simply size for clarity and consistency. The term "event" isn't used consistently across the code, and since the class handles tasks, aligning the terminology would help make the code more intuitive.

src/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+        this.tasks = tasks;
+        eventListSize = tasks.length;
+    }
+
+    public TaskList(TaskList other) {
+        this.tasks = other.tasks;
+        this.eventListSize = other.eventListSize;
+    }
+
+    public void addEvent(String event) {
```
KeerthanaBanukumar: since the class represents tasks, it would be more consistent with the rest of the codebase to rename it to addTask. This would make the code more cohesive and aligned with the naming conventions in the rest of the class.

src/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+            return;
+        }
+        for (int i = 0; i < eventListSize; i++) {
+            System.out.println((i + 1) + ". " +
+                    (tasks[i].getEventStatus() ? "[X] " : "[ ] ")
+                            + tasks[i].getEventName());
+        }
+    }
+
+    public void markDone(String eventName){
```
KeerthanaBanukumar: These methods are quite long due to the multiple actions they perform (checking event status, printing messages, and invoking listTasks). To improve readability, you might want to split these methods into smaller functions. For example, extracting the event status check and the printing logic into a separate method could make the code cleaner and more modular.

src/Tasks.java
```diff
@@ -0,0 +1,25 @@
+public class Tasks {
+
+    private String eventName;
+    private String eventDate;
+    private boolean isCompleted;
```
KeerthanaBanukumar: The logic for checking whether there are tasks in the list and printing the message is repeated in several methods (e.g., markDone, markUndone, addEvent, etc.). To improve code reusability, this logic should be extracted into a helper method that can be reused across different places.

src/chatBot.java
```diff
@@ -0,0 +1,95 @@
+import java.util.Scanner;
+
+public class chatBot{
```
kinneo: In Java, class names should follow PascalCase. Rename chatBot to ChatBot.

src/Hudou.java
```diff
@@ -0,0 +1,13 @@
+import java.util.Scanner;
+
+public class Hudou {
+    public static void main(String[] args) {
+        chatBot chatBot = new chatBot();
+        String input = null;
+        do{
+            Scanner scanner = new Scanner(System.in);
```
kinneo: You're creating a new Scanner instance inside the loop in Hudou.java. Move it outside the loop and close it when done.

src/TaskList.java
```diff
@@ -0,0 +1,117 @@
+public class TaskList {
+
+    private Tasks[] tasks;
+    private int eventListSize;
+
+    public TaskList() {
+        tasks = new Tasks[100];
```
kinneo: magic number, could use a constant

src/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+        eventListSize = tasks.length;
+    }
+
+    public TaskList(TaskList other) {
+        this.tasks = other.tasks;
+        this.eventListSize = other.eventListSize;
+    }
+
+    public void addEvent(String event) {
+        if (eventListSize >= 100){
```
kinneo: magic number, could use a constant

src/chatBot.java
```diff
@@ -0,0 +1,95 @@
...
+                taskList.listTasks();
+                break;
+            case "bye":
+                endSession();
+                break;
+            case "mark":
+                if (isInteger(inputs[1])){
+                    taskList.markDone(Integer.parseInt(inputs[1]));
+                } else{
+                    int firstSpaceIndex = input.indexOf(" ");
```
kinneo: You're extracting event names using indexOf(" "). Instead, use input.substring(input.indexOf(' ') + 1).trim(), but only after checking input.contains(" ").

src/chatBot.java
```diff
@@ -0,0 +1,95 @@
...
+            Integer.parseInt(str);
+            return true;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
+
+    public void processInputs(String input){
+        String[] inputs = input.split(" ");
+        try{
```
kinneo: good that you are using switch cases instead of multiple if else statments

kinneo (commented):
lgtm keep up the good work

src/main/java/Hudou/task/TaskList.java
```diff
@@ -0,0 +1,191 @@
...
+
+
+    public void addTask(String taskInput, boolean isReadFromFile) {
+        //handle exception of too many events being in the list
+        if (tasks.size() >= MAXTASKCOUNT){
+            System.out.println(tooManyTasksNotifier);
+            return;
+        }
+        //process the first arg to get the type of the task
+        //tasks[taskCounter] = classifyTaskTypes(taskInput);
```
aditichadha1310: You might want to remove commented out code.

aditichadha1310 (commented):
overall, good work!

PR #38, opened 2025-02-02
[jwyk] iP

jwyk:


src/main/java/Bob.java
```diff
@@ -0,0 +1,113 @@
+import java.util.Scanner;
+
+public class Bob {
+    private static int current;
+    private static Task[] list;
+
+    //Initializer for Bob
+    public Bob() {
+        current = 0;
```
LeThiHongMinh: The variable 'current' is used to track the number of tasks. It would be clearer if it were renamed to something more descriptive, like 'taskCount', to improve readability and understanding of its role.

src/main/java/Bob.java
```diff
@@ -0,0 +1,113 @@
...
+        String command;
+        int position;
+
+        if (taskPosition != -1) {
+            command = input.substring(0, input.indexOf(' '));
+        } else {
+            command = input;
+        }
+
+        switch (command) {
```
LeThiHongMinh: I like how you use "switch" statement for handling different commands, making it clear and easy to extend if new commands need to be added in the future.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
```
LeThiHongMinh: You've used methods like getStatusIcon() to retrieve the task status in a clean, reusable manner, which is a good practice to avoid redundant code.

src/main/java/Bob.java
```diff
@@ -0,0 +1,113 @@
...
+        case "bye":
+            System.out.println("Bye dude, see you soon again!");
+            break;
+        case "list":
+            for (int i = 0; i < current; i++) {
+                System.out.println(i + 1 + ". " + list[i].getTask());
+            }
+            break;
+        case "mark":
```
LeThiHongMinh: The "mark" and "unmark" may follow a similar pattern, which may need refactoring.

src/main/java/Bob.java
```diff
@@ -0,0 +1,113 @@
+import java.util.Scanner;
+
+public class Bob {
+    private static int current;
+    private static Task[] list;
+
+    //Initializer for Bob
+    public Bob() {
+        current = 0;
+        list = new Task[100];
```
LeThiHongMinh: It's good practice to replace "100" with a named constant to indicate its purpose such as "MAX_TASKS"

LeThiHongMinh (commented):
Great work overall! 👍

src/main/java/Bob.java
```diff
@@ -0,0 +1,114 @@
...
+            String endDate = startDate.substring(endPosition);
+            startDate = startDate.substring(0, endPosition - 2);
+            input = input.substring(taskPosition + 1, input.indexOf('/') - 1);
+            list.add(new Event(input, startDate, endDate));
+            System.out.println("added: " + list.get(current).getTask());
+            current++;
+            System.out.println("Current number of tasks: " + current);
+            break;
+
+        case "todo":
```
melee01: Better to have todo as its own class like how you separate it in Deadline and Event, if there is any additional guideline on modifying ToDo, your current code will need to be changed a lot.

src/main/java/Bob.java
```diff
@@ -0,0 +1,114 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Bob {
+    private static int current;
+    private static ArrayList<Task> list;
+
+    //Initializer for Bob
+    public Bob() {
+        current = 0;
```
melee01: Current variable is indicated to tell on how many tassks are the in the list, however, it is best to have a meaningful definition to every variable you used, it will improve readability, especially when you reviewed and go through your own code, you can grasp and make changes easier.

src/main/java/Bob.java
```diff
@@ -0,0 +1,114 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
```
melee01: I think you can add another layering in interface, such as home page. This will help in case when another additional feature that is outside of this Bob chatbot functionality, where it works alongside this Bob.java, the code will be easier for extension later on. So, can create something like home page, and then call Bob.java inside it. It is neater as well for me.

src/main/java/Bob.java
```diff
@@ -0,0 +1,114 @@
...
+        String command;
+        int position;
+
+        if (taskPosition != -1) {
+            command = input.substring(0, input.indexOf(' '));
+        } else {
+            command = input;
+        }
+
+        switch (command) {
```
melee01: Using Switch and differentiate inputs by cases is smart here, I appreciate it so much.

melee01 (commented):
Overall, it is written well although some improvements can be made.

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+            System.out.println("Current number of tasks: " + taskList.size());
+            Storage.save(taskList);
+            break;
+
+        default:
+            //None of the above commands met
+            System.out.println("Sorry, I didn't get that. Will come back to you...");
+        }
+        System.out.println("____________________________________________________________");
+    }
```
Paulifyer: I think you can shorten this method by breaking up each case into individual methods?

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+     * String constants for each command type
+     */
+    private static final String COMMAND_BYE = "bye";
+    private static final String COMMAND_LIST = "list";
+    private static final String COMMAND_MARK = "mark";
+    private static final String COMMAND_UNMARK = "unmark";
+    private static final String COMMAND_DEADLINE = "deadline";
+    private static final String COMMAND_TODO = "todo";
+    private static final String COMMAND_EVENT = "event";
+    private static final String COMMAND_DELETE = "delete";
```
Paulifyer: Instead of using strings here you could maybe use an enumerator?

src/main/java/bob/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+        return f.exists();
+    }
+
+    /**
+     * Saves the list into a text file, each separated by 1 line
+     *
+     * @param list ArrayList containing Tasks
+     */
+
+    public static void save(ArrayList<Task> list) throws BobException, IOException {
```
Paulifyer: Technically you do not need to throw IOException here as your code already catches it

src/main/java/bob/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+
+public class Storage {
+    private static final String filePath = "bob.txt";
+    private static final String DIV = "|";
+
+    /**
+     * Returns whether a save file exists or not
+     */
+
+    public static boolean exists() {
```
Paulifyer: The naming of this method could be more clear

PR #43, opened 2025-02-02
[Choy Wayne] iP

WayneCh0y:


src/main/java/Printer.java
```diff
@@ -0,0 +1,51 @@
+import java.util.List;
+
+public class Printer {
+    private static final String LINE_SEPARATER = "____________________________________________________________";
```
jxromy: Good use of a constant to avoid magic strings!

jxromy:
```suggestion
    private static final String LINE_SEPARATOR = "____________________________________________________________";
```

src/main/java/Printer.java
```diff
@@ -0,0 +1,51 @@
...
+        printLine();
+    }
+
+    public static void end() {
+        System.out.println("Bye! Don't miss me too much, okay?");
+        printLine();
+    }
+
+    public static void printLine() {
+        System.out.println(LINE_SEPARATER);
```
jxromy:
```suggestion
        System.out.println(LINE_SEPARATOR);
```

src/main/java/Printer.java
```diff
@@ -0,0 +1,51 @@
...
+        printEntry(task);
+        printLine();
+    }
+
+    public static void printTaskAdded(int listSize, Task eventTask) {
+        Printer.printLine();
+        System.out.println("Got it! Task added.");
+        System.out.println(eventTask);
+        System.out.println("Now you have " + listSize + " tasks in the list.");
+        Printer.printLine();
```
jxromy: Perhaps create helper method printWithLines(String message) to avoid repeating printLine() calls

src/main/java/Printer.java
```diff
@@ -0,0 +1,51 @@
...
+        System.out.println(LINE_SEPARATER);
+    }
+
+    public static void printIndexNumber(int index) {
+        System.out.print((index + 1) + ". ");
+    }
+
+    public static void printEntry(Task list) {
+        System.out.println(list);
+    }
```
jxromy: Not sure if "list" is appropriate here as it suggests that a collection is being passed, perhaps replace it with "task"

src/main/java/Gojo.java
```diff
@@ -0,0 +1,45 @@
+import java.util.Scanner;
+
+public class Gojo {
+
+    private static String getTaskType(String line, Scanner in) {
+        return (line.split(" ", 2))[0];
+    }
+
+    private static void start(Scanner in, TaskManager taskManager) {
```
jxromy: I like how the start() method efficiently handles user input in a while-loop with a switch statement 👍

src/main/java/Gojo.java
```diff
@@ -0,0 +1,45 @@
+import java.util.Scanner;
+
+public class Gojo {
+
+    private static String getTaskType(String line, Scanner in) {
+        return (line.split(" ", 2))[0];
+    }
```
jxromy: Scanner in is not being used in getTaskType()

src/main/java/Gojo.java
```diff
@@ -0,0 +1,45 @@
...
+        String line;
+        while(true) {
+            line = in.nextLine();
+            String taskType = getTaskType(line, in);
+            switch (taskType) {
+            case "bye":
+                return;
+            case "list":
+                taskManager.printTaskList();
+                continue;
```
jxromy: Perhaps use break instead of continue as continue is redundant inside a while loop

src/main/java/Gojo.java
```diff
@@ -0,0 +1,45 @@
...
+            case "event":
+                taskManager.event(line);
+                continue;
+            case "mark":
+                taskManager.mark(line);
+                continue;
+            case "unmark":
+                taskManager.unmark(line);
+            }
```
jxromy, line 57:
```suggestion
                break;
            default: 
                System.out.println("Sorry, I don't understand that command.");
            }
```

src/main/java/Gojo.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    public static void main(String[] args) {
+        TaskManager taskManager = new TaskManager();
+        Scanner in = new Scanner(System.in);
```
jxromy: Should the scanner be closed? Perhaps add a in.close() to avoid resource leaks

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public static Task getTaskAtIndex(int index) {
+        return taskList.get(index);
+    }
+
+    public static int getListSize() {
+        return taskList.size();
+    }
+
+    public static void printTaskList() {
```
jxromy: Perhaps use an if block in printTaskList to handle empty lists

jxromy (changes requested):
Looks great! Just a few nits to fix and exceptions to handle

src/main/java/Gojo.java
```diff
@@ -0,0 +1,51 @@
...
+        return (line.split(" ", 2))[0];
+    }
+
+    private static void start(Scanner in, TaskManager taskManager) {
+        String line;
+        while(true) {
+            line = in.nextLine();
+            String taskType = getTaskType(line);
+            switch (taskType) {
+            case "bye":
```
TVageesan: The strings "bye", "list" etc. are magic strings. Consider using constants or a enum to represent the strings instead.

src/main/java/Managers/TaskManager.java
```diff
@@ -0,0 +1,199 @@
...
+
+public class TaskManager {
+    private static List<Task> taskList;
+
+    private static void isValidDeadlineCommand(String command) throws IllegalCommandException {
+        String[] words = command.split(" ", 2);
+        if (words.length < 2) {
+            throw new IllegalCommandException();
+        }
+        if (words[1].split("\\s*/by\\s*").length < 2) {
```
TVageesan: Good idea to use regex for the string comparisons, but you might want to consider putting them in constants for better readability

src/main/java/Managers/TaskManager.java
```diff
@@ -0,0 +1,199 @@
...
+            throw new IndexOutOfBoundsException();
+        }
+    }
+
+    private static void isValidEventCommand(String command) throws IllegalCommandException {
+        String[] words = command.split(" ", 2);
+        if (words.length < 2) {
+            throw new IllegalCommandException();
+        }
+        if (words[1].split("\\s*/from\\s*|\\s*/to\\s*").length < 3) {
```
TVageesan: It would also help readability to assign each part of the words arrat to a appropriate variable before doing logic. For example, here could be `argString = words[1]; args = argString.split(EVENT_ARG_PATTERN); if args.length < 3...` 
This way someone can understand the if condition at a glance without fully understanding what words is or what words[1] is. You can apply this principle in many parts of your code for better readability, not just this function.

TVageesan (commented):
Overall well done, I like how you went the extra mile to abstract functionality such as making a Printer class. Made some points about readability, but logic wise looks good

PR #45, opened 2025-02-02
[icknee] iP

icknee:


src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+        Scanner in = new Scanner(System.in);
+        Task[] taskList = new Task[MAX_TASKS];
+        int taskCount = 0;
+
+        line = in.nextLine();
+        String command = line.split(" ")[0];
+
+
+        while (!line.equals("bye")) {
+            switch (command) {
```
TayGuangSheng: Switch and case should have the same indentation

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+public class Bird {
+    private static final int MAX_TASKS = 100;
+    private static final String INDENT = "    ";
+
+    public static void printReply(String message) {
+        System.out.println(INDENT + message);
+    }
+
+    public static void printGreeting() {
+        printLine();
```
TayGuangSheng: You can reduce the number of printLine() calls to improve readability.

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+                case "list":
+                    printTaskList(taskList, taskCount);
+                    break;
+                case "mark":
+                case "unmark":
+                    try {
+                        toggleTaskStatus(taskList, Integer.parseInt(line.split(" ")[1]), command.equals("mark"));
+                    } catch (NullPointerException | NumberFormatException | ArrayIndexOutOfBoundsException e) {
+                        printLine();
+                        printReply("Invalid task number");
```
TayGuangSheng: I think you can also let the user know what is the valid task number in this case.

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+    public static void addTask(Task[] taskList, String task, int taskCount) {
+        printLine();
+        printReply("added: " + task);
+        printLine();
+        Task newTask = new Task(task);
+        taskList[taskCount] = newTask;
+    }
+
+
+    public static void toggleTaskStatus(Task[] taskList, int taskNumber, boolean done) {
```
TayGuangSheng: You can consider splitting it into 2 methods: [1] markTaskDone() and markTaskUndone() to improve readability

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+    public static void printLine() {
+        System.out.println(INDENT + "____________________________________________________________");
+    }
+
+    public static void printGoodBye() {
+        printLine();
+        printReply("Bye. Hope to see you again soon!");
+        printLine();
+    }
+
```
TayGuangSheng: Try to remain consistent in the number of lines between code. In this case you have 2 lines here as compared to 1 for other parts of the code

TayGuangSheng (commented):
Generally, this code is well written. If possible, you can spend some time to configure the anomaly cases.

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+        String line;
+        Scanner in = new Scanner(System.in);
+        Task[] taskList = new Task[MAX_TASKS];
+        int taskCount = 0;
+
+        line = in.nextLine();
+        String command = line.split(" ")[0];
+
+
+        while (!line.equals("bye")) {
```
JudHoka: I think this while loop visibly still have the arrow head structure, try to group some of them necessarily into methods to avoid this

src/main/java/bird.java
```diff
@@ -2,54 +2,87 @@
...
+
+    public static void markTask(Task[] taskList, int taskNumber, int taskCount) {
+        if (taskNumber > taskCount) {
+            printLine();
+            reply("Invalid task number");
+            printLine();
+            return;
+        }
+        taskList[taskNumber - 1].setDone(true);
+        printLine();
```
JudHoka: I've noticed you also tried doing SLAP by implementing print methods, great job!

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
+import java.util.Scanner;
+
+
+public class Bird {
+    private static final int MAX_TASKS = 100;
+    private static final String INDENT = "    ";
+
+    public static void printReply(String message) {
```
JudHoka: The methods you made is all below 30 LoC, awesome!

src/main/java/Bird.java
```diff
@@ -0,0 +1,96 @@
...
+        printLine();
+        Task newTask = new Task(task);
+        taskList[taskCount] = newTask;
+    }
+
+
+    public static void toggleTaskStatus(Task[] taskList, int taskNumber, boolean done) {
+        taskList[taskNumber - 1].setDone(done);
+        printLine();
+        if (done) {
```
JudHoka: And by far no complicated expressions in if statements, very good

src/main/java/bird/Parser.java
```diff
@@ -0,0 +1,109 @@
...
+            c = new ListCommand();
+            break;
+        case "mark", "unmark":
+            int markTaskNumber;
+            try {
+                markTaskNumber = Integer.parseInt(fullCommand.split(" ")[1]);
+            } catch (Exception e) {
+                throw new InvalidCommandException("Please enter a valid task number");
+            }
+            c = new MarkCommand((action.equals("mark")), markTaskNumber);
```
yijiano: Perhaps the check for invalid command inputs can be refactored in a way such that its handle by the command class instead?

yijiano (commented):
Good work. Just one small suggestion in my review.

src/main/java/bird/Parser.java
icknee: Is this for mark case only or for all the other cases also?

PR #46, opened 2025-02-02
[Pang Zixi] iP

HalFentise:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task{
+    private String deadline;
+
+    Deadline(String taskName,String deadline) {
```
NandhithaShree: A space after commas are usually more readable

src/main/java/Special.java
```diff
@@ -0,0 +1,12 @@
+public class Special {
```
NandhithaShree: It's great you have a class like this so that you have less repetition within your code! Goodjob :)

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,99 @@
...
+            String time = deadline.split(" /by ")[1];
+            task = new Deadline(name, time);
+        } else if (type.equals("event")) {
+            String event = parts[1];
+            String name = event.split(" /from ")[0];
+            String time = event.split(" /from ",2)[1];
+            String startTime = time.split(" /to ")[0];
+            String endTime = time.split(" /to ",2)[1];
+            task = new Event(name, startTime, endTime);
+        } else {
```
NandhithaShree: It is awesome you have ensured that exceptions are handled

src/main/java/Todo.java
```diff
@@ -0,0 +1,10 @@
+public class Todo extends Task{
```
NandhithaShree: ToDo would be the correct way to name this class, instead of Todo

NandhithaShree (commented):
Generally well done!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+        super(taskName);
+        this.deadline = deadline;
+    }
+
+    public String toString() {
+        String state = isDone ? "[X]" : "[ ]";
+        return "[D]" + state + " " + taskName + " (by:" + deadline + " )";
+    }
```
minhbgn: It is nice that you create a variable state to change isDone to a String. I suggest you create another String variable, description, for example, to store the whole ( "[D]" + state + " " + taskName + " (by:" + deadline + " )" ) before returning. I think it can explain what the return statement is returning, and reduce the complexity of the return statement. If you decide to follow, make sure to check other class with the same situation as well.

src/main/java/Special.java
```diff
@@ -0,0 +1,12 @@
...
+    static void greeting() {
+        System.out.println("--------------------------------");
```
minhbgn: I suggest create a constant: final String LINE = "-------------------------------", so it will be less tedious typing the same thing over and over again.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,99 @@
...
+            String name = deadline.split(" /by ")[0];
+            String time = deadline.split(" /by ")[1];
+            task = new Deadline(name, time);
+        } else if (type.equals("event")) {
+            String event = parts[1];
+            String name = event.split(" /from ")[0];
+            String time = event.split(" /from ",2)[1];
+            String startTime = time.split(" /to ")[0];
+            String endTime = time.split(" /to ",2)[1];
+            task = new Event(name, startTime, endTime);
```
minhbgn: I wonder if you can create more specific methods, such as addToDo, addDealine, addEvent, etc. to make this add method less long?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            System.out.println("Sorry, I cannot recognize this task :(\n" +
+                            "here are the types I supported:\ntodo, deadline, event");
+            System.out.println("--------------------------------");
+            return;
+        }
+
+        //add to list
+        taskList = Arrays.copyOf(taskList,taskList.length + 1);
```
minhbgn: You can try checking out the ArrayList class to see if it's easier to use than Arrays because it can dynamically add items in, like a list in python.

minhbgn (commented):
Overall, good job! You had followed the conventions closely and gave me a very hard time finding the violations in following Code Quality.

src/main/java/Parser.java
```diff
@@ -0,0 +1,81 @@
+import Task.*;
+import Exceptions.*;
+import java.io.*;
+
+public class Parser {
+    public static void parse(String input, TaskList taskList, Ui ui, Storage storage) {
+        String[] parts = input.split("\\s+", 2);
```
aditichadha1310, line 22: Avoid usage of Magic strings

src/main/java/Parser.java
```diff
@@ -0,0 +1,81 @@
...
+                case "todo":
+                    if (parts.length < 2) {
+                        throw new TaskException("The description of a todo cannot be empty.");
+                    }
+                    Task todo = new Todo(parts[1]);
+                    taskList.addTask(todo);
+                    storage.saveTasks(taskList.getTasks());
+                    ui.showTaskAdded(todo, taskList.getTasks().size());
+                    break;
+                case "event":
```
aditichadha1310: Avoid arrow head style code

src/main/java/Parser.java
```diff
@@ -0,0 +1,81 @@
+import Task.*;
+import Exceptions.*;
+import java.io.*;
+
+public class Parser {
```
aditichadha1310, line 9: Please try to keep the method to about 30 lines of code. Avoid lengthy methods

PR #48, opened 2025-02-02
[Jahnavi Pydi] iP

pjahn31:


src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Emily {
+    public static ArrayList<Task> tasks = new ArrayList<>();
```
lakshm1i: I noticed you're using ArrayList directly here. Would using the interface type List<Task> instead of ArrayList<Task> make it easier to switch to another list implementation in the future?

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
+public class Task {
+    protected String description;
+    protected Boolean isDone;
```
lakshm1i: Would it be better to use the primitive type boolean instead of Boolean? It avoids unnecessary object creation and aligns with Java conventions for simple true/false flags.

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
+public class Task {
+    protected String description;
```
lakshm1i: Should the description and isDone fields be private instead of protected to fully encapsulate the data? It might prevent unintended modifications from outside the class.

src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
...
+            System.out.println("  [X] " + task.description);
+        }
+    }
+
+    public static void unmarkTask(int taskNumber) {
+        if (taskNumber > 0 && taskNumber <= tasks.size()) {
+            Task task = tasks.get(taskNumber - 1);
+            task.unmarkAsDone();
+            System.out.println("Ok, I've unmarked this task as not done yet:");
+            System.out.println("  [ ] " + task.description);
```
lakshm1i: Would it be better to access description via a getter method instead of direct field access for consistency with encapsulation principles?

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
...
+    protected Boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); //mark done task with X
```
lakshm1i: I like how concise this is!

lakshm1i (commented):
Great job!

src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Emily {
+    public static ArrayList<Task> tasks = new ArrayList<>();
+
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+        System.out.println("Hello! I'm Emily.");
```
adoorknob: maybe you could abstract this out into a 'welcome message' method for readability!

src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
...
+                break;
+            } else if (input.equalsIgnoreCase("list")) {
+                printTasks();
+            } else if (input.startsWith("mark ")) {
+                markTaskAsDone(Integer.parseInt(input.substring(5)));
+            } else if (input.startsWith("unmark ")) {
+                unmarkTask(Integer.parseInt(input.substring(7)));
+            } else {
+                addTask(input);
+            }
```
adoorknob: maybe you could use a switch statement instead, for better scalability!
for example,
```
command = input.toLowercase();
switch (command) {
  case "xxx": // ...
  case "yyy": // ...
}
```

src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static void printTasks() {
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < tasks.size(); i++) {
+            Task task = tasks.get(i);
+            System.out.println((i + 1) + ".[" + task.getStatusIcon() + "] " + task.description);
+        }
+    }
+
+    public static void markTaskAsDone(int taskNumber) {
```
adoorknob: the name of this method is a bit different from the `unmarkTask` function, which might be a bit confusing

src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
...
+            task.markAsDone();
+            System.out.println("Nice! I've marked this task as done: ");
+            System.out.println("  [X] " + task.description);
+        }
+    }
+
+    public static void unmarkTask(int taskNumber) {
+        if (taskNumber > 0 && taskNumber <= tasks.size()) {
+            Task task = tasks.get(taskNumber - 1);
+            task.unmarkAsDone();
```
adoorknob: i like the abstraction of this into another method, really upholding the OOP spirit

src/main/java/Task.java
Markneoneo: Using boolean is better than Boolean in Java because boolean is a primitive type, while Boolean is an object wrapper class.

src/main/java/Emily.java
```diff
@@ -0,0 +1,61 @@
...
+        if (taskNumber > 0 && taskNumber <= tasks.size()) {
+            Task task = tasks.get(taskNumber - 1);
+            task.markAsDone();
+            System.out.println("Nice! I've marked this task as done: ");
+            System.out.println("  [X] " + task.description);
+        }
+    }
+
+    public static void unmarkTask(int taskNumber) {
+        if (taskNumber > 0 && taskNumber <= tasks.size()) {
```
adoorknob: nice catch for the task number! maybe consider making this throw an exception instead, so the user knows why the command didn't work

adoorknob (commented):
lgtm apart from some nits, i like your naming and the readability of your code!

src/main/java/Emily.java
```diff
@@ -2,42 +2,60 @@
...
+            } else if (input.startsWith("mark ")) {
+                markTaskAsDone(Integer.parseInt(input.substring(5)));
+            } else if (input.startsWith("unmark ")) {
+                unmarkTask(Integer.parseInt(input.substring(7)));
             } else {
-                System.out.println("Sorry I do not know what you are saying. Please type again!");
+                addTask(input);
             }
         }
-        sc.close();
```
Markneoneo: Maybe use switch cases instead of many if else statements. The code will be clearer and more structured.

src/main/java/Emily.java
```diff
@@ -2,42 +2,60 @@
...
+            Task task = tasks.get(i);
+            System.out.println((i + 1) + ".[" + task.getStatusIcon() + "] " + task.description);
         }
     }
 
-    public static void addTask(String task) {
-        tasks.add(task);
-        System.out.println("added: " + task);
+    public static void markTaskAsDone(int taskNumber) {
+        if (taskNumber > 0 && taskNumber <= tasks.size()) {
```
Markneoneo: The code checks if taskNumber is within range but does not handle invalid inputs.
Fix: Add an else block to notify the user when an invalid task number is provided.
        System.out.println("⚠ Error: Invalid task number.");

src/main/java/Emily.java
```diff
@@ -2,42 +2,60 @@
...
+            Task task = tasks.get(taskNumber - 1);
+            task.markAsDone();
+            System.out.println("Nice! I've marked this task as done: ");
+            System.out.println("  [X] " + task.description);
+        }
+    }
+
+    public static void unmarkTask(int taskNumber) {
```
Markneoneo: Method Naming Consistency:
markTaskAsDone and unmarkTask have inconsistent naming styles

src/main/java/Emily.java
```diff
@@ -2,42 +2,60 @@
 import java.util.ArrayList;
 
 public class Emily {
-    public static ArrayList<String> tasks = new ArrayList<>();
+    public static ArrayList<Task> tasks = new ArrayList<>();
```
Markneoneo: Nice implementation of the new Class Task.

Markneoneo (commented):
The code has syntax issues (missing closing brace, inconsistent indentation) and readability concerns (redundant print statements, unclear indexing). It lacks proper error handling for invalid task numbers and may risk a NullPointerException if tasks is uninitialized. Naming conventions are inconsistent, and duplicate logic can be extracted into helper methods for maintainability.

src/main/java/Emily.java
```diff
@@ -0,0 +1,69 @@
...
+                addTask(new Deadline(parts[0], parts[1]));
+            } else if (input.startsWith("event ")) {
+                String[] parts = input.substring(6).split(" from | to ");
+                addTask(new Event(parts[0], parts[1], parts[2]));
+            } else {
+                System.out.println("Sorry, I am not sure what that means.");
+            }
+        }
+    }
+
```
Alfred-Goh02: Consider moving these methods to another class, and separate print lines to another class as well. This is for easier management and to ensure that each file is representative of what methods it contains.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
...
+    protected String by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
```
Alfred-Goh02: Avoid using magic strings and assign it to a meaningful constant.

src/main/java/Emily_duke_ref.java
```diff
@@ -1,4 +1,4 @@
-public class Duke {
```
Alfred-Goh02: If this file is used just for your reference, you don't have to commit it.

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Event(String description,String from, String to) {
+        super(description);
+        this.from = from;
+        this.to = to;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + "(from: " + from + " to: " + to + ")";
```
Alfred-Goh02: Magic Strings here as well.

src/main/java/Todo.java
```diff
@@ -0,0 +1,10 @@
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+    }
+
```
Alfred-Goh02: The method is compact and easy to understand, nice!

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
```
Alfred-Goh02: Very concise methods which follow conventions, nice!

PR #50, opened 2025-02-02
[Deanson-Choo] iP

Deanson-Choo:


src/main/java/Reader.java
```diff
@@ -0,0 +1,29 @@
...
+                }
+                else if (line.equalsIgnoreCase("list")) { //Case 2: List all items
+                    taskManager.listTasks();
+                }
+                else if (line.toLowerCase().startsWith("mark") || line.toLowerCase().startsWith("unmark")) { //Case 3: Marking items
+                    taskManager.updateItemStatus(line.split(" "));
+                }
+                else {
+                    taskManager.addNewItem(line); //Case 4: add items
+                }
```
hannahtay: Is this the correct structure to type out if-else statements?

src/main/java/Reader.java
```diff
@@ -0,0 +1,29 @@
...
+        public void startScanning() {
+            while (true) {
+                String line = scanner.nextLine();
+               if ((line.equalsIgnoreCase("bye"))) { //Case 1: Exit Program
+                    break;
+                }
+                else if (line.equalsIgnoreCase("list")) { //Case 2: List all items
+                    taskManager.listTasks();
+                }
+                else if (line.toLowerCase().startsWith("mark") || line.toLowerCase().startsWith("unmark")) { //Case 3: Marking items
```
hannahtay: This line is too long, perhaps you can separate the conditions?

src/main/java/ToDo.java
```diff
@@ -0,0 +1,9 @@
+public class ToDo extends Task{
```
hannahtay: I like that the classes are separate from each other, makes it easier to read

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
+    protected String name = null;
+    boolean done;
```
hannahtay: Should this be renamed to follow the coding standard for boolean variables?

hannahtay (commented):
Overall code was readable and looks good!
Just a few small changes to be made :)

src/main/java/Reader.java
Deanson-Choo: I'll take note! Thank you!

src/main/java/Task.java
Deanson-Choo: Will take note! Thank you for the comment!

src/main/java/UIHelper.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public static void printError(String errorMessage) {
+        System.out.println("ERROR: " + errorMessage);
+    }
+
+    public static String getSeperator() {
+        return SEPARATOR;
+    }
+}
```
jyukuan: I pretty like the way that you separate all the function in different classes. It is so clear to read.

src/main/java/Reader.java
jyukuan: How about just input "mark" without index? Is it might be an error?

src/main/java/Jeff.java
```diff
@@ -2,91 +2,30 @@
...
         System.out.println(SEPARATOR);
     }
 
-    public static void Intro() {
+    public void intro() {
         printWithSeparator("Hello! I'm Jeff!", "What can I do for you?");
     }
 
-    public static void Outro() {
+    public void outro() {
```
jyukuan: A bit confused to see the function"outro" at first time. Maybe change the function name?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,80 @@
...
+            UIHelper.printWithSeparator("Please input a task number!");
+            return;
+        }
+        int task_index = Integer.parseInt(words[1]) - 1;
+
+        //Error 2
+        if (task_index < 0 || task_index >= taskCount) {
+            UIHelper.printWithSeparator("Please input a valid task number!");
+            return;
+        }
```
jyukuan: So clear about error classifaction !

jyukuan (commented):
I think it was general well written.

src/main/java/jeff/Reader.java
```diff
@@ -0,0 +1,63 @@
...
+
+        private boolean isMarkOrUnmarkCommand(String line) {
+            return line.toLowerCase().startsWith("mark") || line.toLowerCase().startsWith("unmark");
+        }
+
+        private boolean isTaskCommand(String line) {
+            return line.toLowerCase().startsWith("todo") ||
+                    line.toLowerCase().startsWith("deadline") ||
+                    line.toLowerCase().startsWith("event");
+        }
```
TVageesan: There is a common functionality here where for each helper function, you either equalsIgnoreCase or first convert toLowerCase(). Is there a way to abstract this common functionality of lowering the case of the input before comparing it to something?

TVageesan: The string comparisons "bye", "list" etc. are magic strings. Perhaps an enum could be used for the command comparisons instead?

TVageesan: Extra comment (optional, does not affect code quality if you leave as is): Once you've abstracted that common functionality out, is there a need for the helper functions anymore? Can the overall logic for this file be simplified?

src/main/java/jeff/TaskManager.java
```diff
@@ -0,0 +1,108 @@
+package jeff;
+
+public class TaskManager{
+    private final Task[] tasks = new Task[100];
+    private int taskCount = 0;
+
+
+    public void addNewItem(String taskDetails) throws JeffException {
+        String[] parts = taskDetails.split(" ", 2);
+        //part[0] is the task type, //parts[1] is the details
```
TVageesan: Instead of directly passing part[1] as input arguments/use in your code, perhaps you could split parts into two variables taskType, taskDetails. This will make your code easier to parse at first glance instead of having to read this comment and hold the "meanings" of each index in their head as they read.

src/main/java/jeff/TaskManager.java
```diff
@@ -0,0 +1,108 @@
...
+        if (task_index < 0 || task_index >= taskCount) { //Error 3: Invalid Task Number
+            UIHelper.printError("Please input a valid task number!");
+            return;
+        }
+
+        Task currTask = tasks[task_index];
+        boolean isMark = instruction.equalsIgnoreCase("mark");
+        currTask.setStatus(isMark); //if mark - set to true, else unmark - set to false
+
+        String status_message = isMark
```
TVageesan, line 95: The variable name status_message is in snake_case. Perhaps the name statusMessage would better fit our naming convention of camelCase?

TVageesan (commented):
Overall really well done, it's clear you've put effort into abstracting common functionality and writing readable code. Left some comments with nitpicks/things to consider, take a look when you can.

PR #51, opened 2025-02-02
[nandhananm7] ip

nandhananm7:


src/main/java/Talko.java
```diff
@@ -0,0 +1,54 @@
...
+
+        System.out.println("Hello, you can call me");
+        System.out.println(symbol);
+        System.out.println("How can I help you today?");
+        System.out.println(line);
+
+        while (true) {
+            String input = in.nextLine().trim();
+            String command = input.split(" ", 2)[0];
+            String argument = input.contains(" ") ? input.substring(input.indexOf(" ") + 1).trim() : "";
```
minhbgn: I would prefer to write this in if-else statement. I think it can help keep the line of code easier to understand.

src/main/java/Talko.java
```diff
@@ -0,0 +1,54 @@
+import java.util.Scanner;
+
+public class Talko {
+    public static void main(String[] args) {
+        Scanner in = new Scanner(System.in);
+        TaskManager taskManager = new TaskManager();
+
+        String symbol = "+-+-+-+-+-+\n" + "|t|a|l|k|O|\n" + "+-+-+-+-+-+";
+        String line = "----------------------------------------------";
```
minhbgn: I believe these are constants. I would prefer to write the variable name in ALL CAPITAL, i.e, SYMBOL, and LINE.

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public void mark(boolean isDone) {
+        this.isDone = isDone;
+        System.out.println(isDone ? "Marked as Done!" : "Marked as Not Done!");
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
```
minhbgn: I think it would be nice if you can include a variable explaining what "X" and " " are, to avoid "magic string". This would also help reduce the complexity of the return statement.

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public String getDescription() {
+        return "[" + getStatusIcon()+ "] " + description;
+    }
+}
```
minhbgn: I would also suggest to create a String variable with a name, then return the String. It might explain better what the return statement is returning.

minhbgn (commented):
Overall, I like how you follow the coding standard closely. It is very nice that I had to nit pick small details to give the review comment.

src/main/java/Talko.java
nandhananm7: Thanks for pointing all the magic strings out! I will resolve them

src/main/java/Talko.java
```diff
@@ -0,0 +1,54 @@
...
+        System.out.println(symbol);
+        System.out.println("How can I help you today?");
+        System.out.println(line);
+
+        while (true) {
+            String input = in.nextLine().trim();
+            String command = input.split(" ", 2)[0];
+            String argument = input.contains(" ") ? input.substring(input.indexOf(" ") + 1).trim() : "";
+
+            switch (command.toLowerCase()) {
```
ElonKoh: I like your use of switch here, making the code logic incredibly legible

src/main/java/Talko.java
ElonKoh: I like that you abstracted these formatting variables, feel free to disagree, but would it act better as a class field that could be accessible by other methods?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,37 @@
...
+    public TaskManager() {
+        this.tasks = new Task[100];
+        this.taskCount = 0;
+    }
+
+    public void addTask(Task task) {
+        if (taskCount < tasks.length) {
+            tasks[taskCount] = task;
+            taskCount++;
+            System.out.println("Added: " + task.getDescription());
```
ElonKoh: I like that you put print statements in your specific class methods, it really tidies up the main method and keeps unecessary lines out of the higher level logic.

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    //constructor
```
ElonKoh: If I were to nitpick, think this comment might not be necessary

src/main/java/Todo.java
```diff
@@ -0,0 +1,10 @@
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String getDescription() {
+        return "[T]" + super.getDescription();
```
ElonKoh: I love your use of inheritance here, keeping the implementation simple and having no repeated code

src/main/java/Talko.java
```diff
@@ -0,0 +1,54 @@
...
+            case "list":
+                taskManager.listTasks();
+                break;
+            case "mark":
+                int markIndex = Integer.parseInt(argument) - 1;
+                taskManager.markTask(markIndex, true);
+                break;
+            case "unmark":
+                int unmarkIndex = Integer.parseInt(argument) - 1;
+                taskManager.markTask(unmarkIndex, false);
```
ElonKoh: I like how you adopt a "switch" logic, using the same method for both the mark and umark command, makes it simpler to read and keeps code repetition to a minimum

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task{
+    private String by;
+
+    public Deadline(String description, String by){
```
ElonKoh: it might be cleaner if you keep the space before "{" consistent throughout the code

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task{
```
ElonKoh: it might be cleaner if you keep the space before "{" consistent throughout the code

ElonKoh (commented):
Overall incredibly well done in making the code legible and straightforward, abstracting methods with little code repetition. I have added some minor notes on consistent formatting, but I am really nitpicking there. Great work!

src/main/java/commands/CommandHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import tasks.Todo;
+
+public class CommandHandler {
+    private TaskManager taskManager;
+
+    public CommandHandler(TaskManager taskManager) {
+        this.taskManager = taskManager;
+    }
+
+    public void handleCommand(String input) throws TalkoException {
```
aditichadha1310, line 26: Try to keep the method length within 30 lines of code.

src/main/java/commands/CommandHandler.java
```diff
@@ -0,0 +1,59 @@
...
+            int unmarkIndex = Integer.parseInt(argument) - 1;
+            taskManager.markTask(unmarkIndex, false);
+            break;
+        case "todo":
+            if (argument.trim().isEmpty()) throw new TalkoException("Description cannot be empty! Please enter a description of your task!");
+            taskManager.addTask(new Todo(argument));
+            break;
+        case "deadline":
+            String[] deadlineDetails = argument.split(" /by ", 2);
+            if (deadlineDetails.length < 2) throw new TalkoException("Oops Invalid format :( Try rewriting it to this format: deadline <description> /by <due date>");
```
aditichadha1310: Keep every line of code within 120 characters.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,105 @@
...
+                    task.mark(true);
+                }
+
+                tasks.add(task);
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println("File not found. Starting with an empty task list.");
+        }
+
+        return tasks;
```
aditichadha1310, line 86: try to avoid the arrow head style code.

PR #52, opened 2025-02-02
[Tan Zhe Hui] iP

ae-24:


src/main/java/Echo.java
```diff
@@ -1,11 +1,10 @@
 import java.util.Scanner;
 
-
 public class Echo {
     public static void main(String[] args) {
         String line;
         Scanner in = new Scanner(System.in);
-        String[] taskList = new String[100];
+        Task[] taskList = new Task[100];
         int listIndex = 0;
```
zexueteh: Perhaps could add access modifiers e.g. private static

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+        isCompleted = true;
+    }
+
+    public void markAsIncomplete(){
+        isCompleted = false;
+    }
+
+    public void printTask() {
+        System.out.println("[" + getStatusIcon() + "] " + description);
+    }
```
zexueteh: Overriding toString may be a cleaner and more extendable way to print Tasks. This allows for polymorphism in the future.

src/main/java/Echo.java
zexueteh: nit: The naming of listIndex could be more descriptive. Perhaps numTask or taskCounts would convey the attribute meaning more clearly.

src/main/java/Echo.java
```diff
@@ -1,11 +1,10 @@
...
 
         System.out.println("____________________________________________________________");
```
zexueteh: Since this line separator is used multiple times, this could be stored as a constant variable like LINE_SEPERATOR to avoid repetition?

src/main/java/Echo.java
```diff
@@ -25,16 +24,35 @@ public static void main(String[] args) {
...
                 } else {
                     for (int i = 0; i < listIndex; i++) {
-                        System.out.println(taskList[i]);
+                        System.out.print((i + 1) + ".");
+                        taskList[i].printTask();
                     }
                 }
                 System.out.println("____________________________________________________________");
                 continue;
+            } else if (line.toLowerCase().startsWith("mark")) {
```
zexueteh: I like how you cast the user command into lowercase, this validates input allowing for more flexibility!

zexueteh (commented):
Overall, this PR is a great increment from the boilerplate project! However, keep in mind that the Task class is a baseline superclass which will be inherited from. It would greatly benefit scalability to design Task according to OOP Principles. Looking forward to seeing future updates! 🔥🚀

src/main/java/FileManager.java
```diff
@@ -0,0 +1,95 @@
+import Task.*;
```
irving11119: Take note when importing packages, we should try to avoid using wildcard imports. Instead import the specific classes explicitly.

src/main/java/FileManager.java
```diff
@@ -0,0 +1,95 @@
...
+            } else if (line.charAt(line.indexOf("[") + 1) == 'E') {
+                int start = line.indexOf("] ") + 2;
+                int fromIndex = line.indexOf("(from:");
+                int toIndex = line.indexOf("to:");
+
+                String taskDescription = line.substring(start, fromIndex - 1);
+                String fromDate = line.substring(fromIndex + 7, toIndex - 1);
+                String toDate = line.substring(toIndex + 4, line.length() - 1);
+                taskToAdd = new Event(taskDescription, fromDate, toDate);
+                markAndAddTask(line, taskToAdd);
```
irving11119, line 90: While not a huge issue, try looking into how you can make your methods [SLAP](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard)

src/main/java/InputManager.java
```diff
@@ -0,0 +1,59 @@
+import java.util.Scanner;
+import Exception.*;
```
irving11119: Take note to avoid wildcard imports.

src/main/java/FileManager.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+
+        scanner.close();
+
+    }
+
+    private void markAndAddTask(String line, Task taskToAdd) {
+        if (line.charAt(line.indexOf("[") + 4) == 'X') {
```
irving11119, line 106: This expression is a bit complex, try to split it up to make it more readable

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,155 @@
+import Task.*;
+import Exception.*;
```
irving11119: As stated above, avoid wildcard imports

irving11119 (commented):
Overall your code is readable and mostly adheres to the coding standards. Take note of some of the feedback I have left and make the changes as you see fit.

PR #53, opened 2025-02-02
[Yingying] ip

yyingg-243:


src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
...
+        int inputCount = 0;
+
+        System.out.println(LINE_SEPARATOR);
+        System.out.println("Hello! I'm ChattyDuke");
+        System.out.println("What can I do for you?");
+        System.out.println(LINE_SEPARATOR);
+        System.out.println();
+
+        Scanner scanner = new Scanner(System.in);
+        Boolean isBye = false;
```
tejaskumar0: Good naming convention for boolean variables

src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
...
+
+public class ChattyDuke {
+
+    public static final String LINE_SEPARATOR = "____________________________________________________________";
+    public static final int MAX_ITEMS = 100;
+    public static final String INDENTATION = "    ";
+
+    public static void main(String[] args) {
+
+        Task[] items = new Task[MAX_ITEMS];
```
tejaskumar0: plural form used to name arrays, good coding standard

src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
+import java.util.Scanner;
+
+public class ChattyDuke {
```
tejaskumar0, line 6: maybe can include more comments to make the flow clearer

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
```
tejaskumar0, line 5: good use of protected fields

src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
...
+        Scanner scanner = new Scanner(System.in);
+        Boolean isBye = false;
+
+        while(!isBye) {
+            String s1 = scanner.nextLine();
+            System.out.println(INDENTATION + LINE_SEPARATOR);
+
+            if(s1.equalsIgnoreCase("bye")) {
+                isBye = true;
+                System.out.println(INDENTATION + "Bye. Hope to see you again soon!");
```
Yikbing: extraction can be done in the bye statement :D

src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
...
+        System.out.println(LINE_SEPARATOR);
+        System.out.println("Hello! I'm ChattyDuke");
+        System.out.println("What can I do for you?");
+        System.out.println(LINE_SEPARATOR);
+        System.out.println();
+
+        Scanner scanner = new Scanner(System.in);
+        Boolean isBye = false;
+
+        while(!isBye) {
```
Yikbing: can make the happy path prominent by restructuring the code such that the isBye is nested instead of the "usual" case(which is when the user does not type "bye")

src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
...
+            String s1 = scanner.nextLine();
+            System.out.println(INDENTATION + LINE_SEPARATOR);
+
+            if(s1.equalsIgnoreCase("bye")) {
+                isBye = true;
+                System.out.println(INDENTATION + "Bye. Hope to see you again soon!");
+                System.out.println(INDENTATION + LINE_SEPARATOR);
+            } else if (s1.equalsIgnoreCase("list")) {
```
Yikbing: good job for using the else-if statements, avoiding the arrowhead style code!

src/main/java/ChattyDuke.java
```diff
@@ -0,0 +1,83 @@
+import java.util.Scanner;
+
+public class ChattyDuke {
+
+    public static final String LINE_SEPARATOR = "____________________________________________________________";
+    public static final int MAX_ITEMS = 100;
```
Yikbing: good job for explaining the meaning of the 100 and not having "magic numbers" in your code! :D

Yikbing (commented):
Overall, well done on the code! You just need a little bit of extraction of code to be able to let readers know what the code does! :D

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,213 @@
...
+
+    public static void handleNewTask(TaskCommand taskCommand, String s1, Task[] taskLists){
+
+        try{
+            String[] taskInfo = s1.split(" ",2);
+
+            if(taskInfo.length < 2 || taskInfo[1].trim().isEmpty()){
+                throw new ChattyDukeException("Hmmmm, Please include the task description");
+
+            }else{
```
TVageesan: We could reduce nesting here by removing the else. If the condition passes and the error throws, the rest of the code won't execute anyway, so the else is not needed. This is a good example of using the inversion pattern to reduce nesting in general!

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,213 @@
...
+                case TODO:
+                    todoTask(s1, taskLists);
+                    break;
+
+                case DEADLINE:
+                    deadlineTask(s1, taskLists);
+                    break;
+
+                case EVENT:
+                    eventTask(s1, taskLists);
```
TVageesan: The method names 'todoTask', 'deadlineTask', etc. are a bit vague. They do not clearly describe the function's actual purpose - to create a new task object. A general rule of thumb is to have a action/verb in your method names: i.e. `createTodo`, `deleteTodo` etc.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,213 @@
...
+            }else if(fromDescription.length() <= 0){
+                throw new ChattyDukeException("Hmmm! When do you what to start this event?");
+
+            }else if(toDescription.length() <= 0){
+                throw new ChattyDukeException("Hmm! So, when would you like this event to end?");
+            }
+
+
+
+            taskLists[ChattyDuke.inputCount] = new Event(firstPart, fromDescription, toDescription);
```
TVageesan: ChattyDuke's static variable 'inputCount' seems to be keeping track of the number of tasks in your taskList. If the only usage of this variable is within TaskManager, it might make more sense for it to be a instance variable stored only inside TaskManager.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+    }
+
+    public static void todoTask(String s1, Task[] taskLists){
+
+
+        String taskInfo = s1.split(" ", 2)[1];
+        taskLists[ChattyDuke.inputCount] = new Task(taskInfo);
+        Task t = taskLists[ChattyDuke.inputCount];
```
TVageesan: Avoid abbreviatiating variable names that you use multiple places in your code. I promise the 3 characters you save here isn't worth the readability loss for people reading it being used later: `t.markDone()` vs `task.markDone()`.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    public static void eventTask (String s1, Task[] taskLists){
+
+        try{
+            String description = s1.split(" ",2)[1];
+            String[] taskInfo = description.split("/from", 2);
+
+            String firstPart = taskInfo[0].trim();
+            String secondPart = taskInfo[1].trim();
```
TVageesan: 'firstPart', 'secondPart' are not very descriptive variable names. It would be more useful if they reflected their contents, i.e. 'description', 'time'. I noticed this in multiple places.

TVageesan (commented):
Well done logic wise, but you could reconsider some of the variable/method names for better readability!

PR #78, opened 2025-02-03
[QuyDatNguyen] iP

QuyDatNguyen:


src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+            String[] words = text.split(" ");
+            if (text.trim().equals("bye")) {
+                printMessage("Bye. Hope to see you again soon!");
+                break;
+            }
+            else if (text.trim().equals("list")) {
+                System.out.println(line);
+                printTaskList(tasks);
+                System.out.println(line);
+            } else if (words[0].equals("unmark") && words.length == 2 && isValidInteger(words[1])) {
```
ShengBin-101: Perhaps a neater way to handle the user input is to have functions to handle each command type? This would help reduce the amount of if-else nesting and make your main function neater. Furthermore it abstracts away the logic behind the handling of each command.

src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+    public static boolean isValidInteger(String str) {
+        try {
+            Integer.parseInt(str);
+            return true;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
+    public static void main(String[] args) {
+        String line = "=============================";
```
ShengBin-101: Variables that are not meant to be changed during runtime should be changed to constants by declaring them with `final`.

src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+    public static boolean isValidInteger(String str) {
+        try {
+            Integer.parseInt(str);
+            return true;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
```
ShengBin-101: I like how you handle the case when the user inputs a non-integer value. This contributes to your code quality by handling edge cases.

src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+                            + "[" + changeTask.getStatusIcon() + "] " + changeTask.getDescription());
+                }
+            }
+            else {
+                Task t = new Task(text);
+                tasks[taskCount] = t;
+                taskCount++;
+                printMessage("added: " + text);
+            }
+            Scanner nextIn = new Scanner(System.in);
```
ShengBin-101: This line to create another Scanner object may not be necessary. In every iteration of your while loop, a new Scanner instance is created. This may result in unnecessary memory allocation.

Suggestion: Use the existing Scanner in declared at the beginning of main() instead: `text = in.nextLine();`
This way, System.in would be wrapped in only one Scanner instance for the entire program.

ShengBin-101 (commented):
Overall, great work in adhering to the coding standards for naming of variables and functions. However, there are areas to improve with regards to code quality.

More importantly, I think the main loop can be made neater by having functions to handle each different inputs. Perhaps could use a switch case structure to evaluate the command given by the user.

src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+    public static void main(String[] args) {
+        String line = "=============================";
+        System.out.println(line);
+        System.out.println(" /\\_/\\  ");
+        System.out.println("( o.o ) " + " " + "Hello! I'm DanChat");
+        System.out.println(" > ^ <  " + " " + "What can I do for you?");
+        System.out.println(line);
+
+        String text;
+        Task[] tasks = new Task[100];
```
TiangSoonYong: The max size `100` for `tasks `seems like a magic literal. Perhaps it can be declared as a `CONSTANT`? I noticed `String line` and `Scanner in = new Scanner(System.in)` can benefit from this too, as they are repeatedly used.

src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+        int taskCount = 0;
+
+        Scanner in = new Scanner(System.in);
+        text = in.nextLine();
+
+        while (true) {
+            String[] words = text.split(" ");
```
TiangSoonYong: Maybe consider separating `String[] words` into individual strings such as `String commands`, `String arguments` to improve code clarity. `words[i]` is quite ambiguous.

src/main/java/DanChat.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+            else {
+                Task t = new Task(text);
+                tasks[taskCount] = t;
+                taskCount++;
+                printMessage("added: " + text);
+            }
+            Scanner nextIn = new Scanner(System.in);
+            text = nextIn.nextLine();
+        }
```
TiangSoonYong: This if-else block is really well done with its sanity checks! Perhaps the next step would be to extract each cases into individual methods for clearer level of abstraction.

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public String getDescription() {
+        return description;
+    }
+
+    public boolean isDone() {
+        return isDone;
+    }
+
+    public void setDone(boolean done) {
```
TiangSoonYong: > Boolean variables/methods should be named to sound like booleans

Perhaps the naming of `done` can be changed to `isDone`, in order to follow this naming convention

TiangSoonYong (commented):
Overall, I like the code, it is neat and readable. There are only some minor violations and a few optimization for code clarity. All the best!

src/main/java/danchat/ui/DanChat.java
```diff
@@ -0,0 +1,346 @@
+package danchat.ui;
+
+import danchat.exception.*;
```
sevenseasofbri: Avoid using wildcard imports. Import the classes required explicitly.

src/main/java/danchat/ui/DanChat.java
```diff
@@ -0,0 +1,346 @@
...
+import danchat.task.Deadline;
+import danchat.task.Event;
+import danchat.task.Task;
+import danchat.task.Todo;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.File;
+//import java.io.FileWriter;
+
+//import java.lang.reflect.Array;
```
sevenseasofbri: Do not comment out dead code, if it is not required anymore, you can remove it.

sevenseasofbri: + same with line 20, 53

src/main/java/danchat/ui/DanChat.java
```diff
@@ -0,0 +1,346 @@
...
+    private static final String EVENT_BEGIN_DATE_SEPERATOR = " /from ";
+    private static final String EVENT_END_DATE_SEPERATOR = " /to ";
+    private static final String ERROR_MISSING_BEGIN_DATE = "Missing starting date. Please provide starting date after /from";
+    private static final String ERROR_MISSING_END_DATE = "Missing ending date. Please provide ending date after /to";
+    private static final String COMMAND_EVENT_WORD = "event";
+    private static final String COMMAND_EVENT_MESSAGE = "Add new event in your list: ";
+
+    private static final String ERROR_EMPTY_DETAIL = "Details must not be blank";
+    private static final String ERROR_INVALID_INDEX_FORMAT = "Wrong index format. Index must be a positive integer";
+    private static final String COMMAND_DELETE_MESSAGE = "I have deleted this task for you: ";
```
sevenseasofbri: good usage of constants.

src/main/java/danchat/ui/DanChat.java
```diff
@@ -0,0 +1,346 @@
...
+    //    private static Task[] tasks;
+    private static ArrayList<Task> taskList = new ArrayList<>();
+    private static int taskCount;
+
+    public static final String FILE_DIRECTORY = "tasks.txt";
+    private static File file = new File(FILE_DIRECTORY);
+
+    private static final Scanner SCANNER = new Scanner(System.in);
+
+    public static void main(String[] args) {
```
sevenseasofbri: Good job on maintaining a consistent level of abstraction here.

src/main/java/danchat/ui/DanChat.java
```diff
@@ -0,0 +1,346 @@
...
+        String userInput = SCANNER.nextLine();
+        while (userInput.trim().isEmpty()) {
+            userInput = SCANNER.nextLine();
+        }
+        return userInput;
+    }
+
+    private static void executeEventCommand(String detail) throws MissingDateException {
+        String[] splitDetails = detail.split(EVENT_BEGIN_DATE_SEPERATOR, 2);
+        if (splitDetails.length < 2) {
```
sevenseasofbri: Good job making the [happy path prominent](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent).

src/main/java/danchat/ui/DanChat.java
```diff
@@ -0,0 +1,346 @@
...
+            throw new MissingDateException(ERROR_MISSING_BEGIN_DATE);
+        }
+        String description = splitDetails[0].trim();
+
+        String duration = splitDetails[1].trim();
+        String[] splitDurations = duration.split(EVENT_END_DATE_SEPERATOR, 2);
+        if (splitDurations.length < 2) {
+            throw new MissingDateException(ERROR_MISSING_END_DATE);
+        }
+        String from = splitDurations[0].trim();
```
sevenseasofbri: There are many magic literals here it would be helpful to refactor them or add some concise comments on what is happening here.

sevenseasofbri (commented):
Good job so far! I think your mehtods are well formed and concise. However, there are many lines of commented out code, this can hinder readability. Consider removing them.

src/main/java/danchat/ui/DanChat.java
sevenseasofbri: + several other lines

PR #81, opened 2025-02-03
[Betahaxer] ip

Betahaxer:


src/main/java/Bond.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final String TAB = WHITE + "    ";
+    private static final String COMMAND = GREEN + "> ";
+    private static final String UNMARKED = "[ ]";
+    private static final String MARKED = "[X]";
+
+    private static final Task[] taskList = new Task[100];
+    private static int taskCount = 0;
+
+    public static void greet() {
+        String greet = TAB + WHITE + "Hello! I'm Bond, a dog that can predict the future.\n" + TAB + "What can I do for you?\n" + GREEN + COMMAND;
```
LEESY02: Nice class name :D

src/main/java/Bond.java
```diff
@@ -0,0 +1,84 @@
...
+                System.out.print(UNMARKED);
+            }
+            System.out.print(" " + taskList[i].getDescription() + "\n");
+        }
+        System.out.print(GREEN + COMMAND);
+    }
+
+    public static void markTask(int itemNumber) {
+        taskList[itemNumber - 1].markAsDone();
+        System.out.println(WHITE + "Woof! This task was marked as done:");
```
LEESY02: Nit picking, but maybe you could extract all these print lines into a private static method

src/main/java/Bond.java
```diff
@@ -0,0 +1,84 @@
...
+        System.out.println("  " + UNMARKED + " " + taskList[itemNumber - 1].getDescription());
+        System.out.print(COMMAND);
+    }
+
+    public static void main(String[] args) {
+        greet();
+        Scanner in = new Scanner(System.in);
+        String instruction;
+        while (!(instruction = in.nextLine()).equals("bye")) {
+            if (instruction.equals("list")) {
```
LEESY02: Perhaps extracting this into a separate method, and using a case/switch block will be more readable

src/main/java/Bond.java
```diff
@@ -0,0 +1,84 @@
+import java.util.Scanner;
```
LEESY02: **Overall very neat and readable**, but some parts of iP tasks are not completed yet.

src/main/java/Bond.java
```diff
@@ -0,0 +1,84 @@
...
+        System.out.println(TAB + WHITE + "added: " + taskDescription);
+        System.out.print(GREEN + COMMAND);
+    }
+
+    public static void showList() {
+        System.out.println(TAB + "Hmph... The future is uncertain, but these tasks must be completed:");
+        for (int i = 0; taskList[i] != null; i++) {
+            System.out.printf(TAB + WHITE + "%d" + ". ", i + 1);
+            if (taskList[i].getIsDone()) {
+                System.out.print(MARKED);
```
LEESY02: Perhaps you could do this portion in the Task classes' toString() method, by utilising the isDone boolean variable

LEESY02: i.e.:
@Override
public String toString() {
return this.index + MARKED + this.description;
}

index is i + 1 from
System.out.printf(TAB + WHITE + "%d" + ". ", i + 1);
of your code, or you could choose not to have an index variable in your class as well and do it under showList()

LEESY02 (commented):
**Very readable code**, but unfortunately the week4 iP tasks have not been completed yet

src/main/java/Bond.java
Betahaxer: Hey thanks for reviewing, I'll get to the rest of the tasks soon!

src/main/java/Bond.java
Betahaxer: HAHAHA thanks

src/main/java/Bond.java
Betahaxer: That's a good idea and I didn't think of that when I was doing it

src/main/java/Bond.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final Task[] taskList = new Task[100];
+    private static int taskCount = 0;
+
+    public static void greet() {
+        String greet = TAB + WHITE + "Hello! I'm Bond, a dog that can predict the future.\n" + TAB + "What can I do for you?\n" + GREEN + COMMAND;
+        System.out.print(greet);
+    }
+
+    public static void sayBye() {
+        String bye = TAB + WHITE + "Woof. Hope to see you again soon!";
```
je-nnyt: What a great way to say bye!

src/main/java/Bond.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final String YELLOW = "\u001B[33m";
+    private static final String BLUE = "\u001B[34m";
+    private static final String WHITE = "\033[97m";
+    private static final String HIGHLIGHT = "\u001B[43m";
+    private static final String TAB = WHITE + "    ";
+    private static final String COMMAND = GREEN + "> ";
+    private static final String UNMARKED = "[ ]";
+    private static final String MARKED = "[X]";
+
+    private static final Task[] taskList = new Task[100];
```
je-nnyt: Maybe declare a constant for 100 to avoid magics number?

src/main/java/Bond.java
```diff
@@ -0,0 +1,91 @@
...
+                break;
+            default:
+                addToList(input);
+                break;
+            }
+            input = in.nextLine();
+        }
+        sayBye();
+    }
+}
```
je-nnyt: Great layout and good job on following naming conventions!

src/main/java/Bond.java
```diff
@@ -0,0 +1,91 @@
...
+    public static void addToList(String taskDescription) {
+        taskList[taskCount] = new Task(taskDescription);
+        taskCount += 1;
+        System.out.println(TAB + WHITE + "added: " + taskDescription);
+        System.out.print(GREEN + COMMAND);
+    }
+
+    public static void showList() {
+        System.out.println(TAB + "Hmph... The future is uncertain, but these tasks must be completed:");
+        for (int i = 0; taskList[i] != null; i++) {
```
je-nnyt: instead of taskList[i] as the for loop bound, would it be safer to use taskCount?

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsNotDone() {
```
je-nnyt: Just a personal preference but naming it as unmarkAsDone might be more intuitive

je-nnyt (commented):
Overall looks good!

src/main/java/Bond/Bond.java
```diff
@@ -0,0 +1,325 @@
+package Bond;
```
sevenseasofbri, line 1: > From java coding standard: Names representing packages should be in all lower case.

src/main/java/Bond/Bond.java
```diff
@@ -0,0 +1,325 @@
...
+
+    private static ArrayList<Task> tasks = new ArrayList<>();
+
+    public static void greet() {
+        System.out.print(GREETING);
+    }
+
+    public static void sayBye() {
+        System.out.println(GOODBYE);
+    }
```
sevenseasofbri: nice use of constants

src/main/java/Bond/Bond.java
```diff
@@ -0,0 +1,325 @@
...
+            default:
+                System.out.println(TAB + INVALID_MARK_COMMAND);
+                break;
+            }
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println(TAB + INVALID_TASK_NUMBER);
+        }
+    }
+
+    private static void findAndMarkTask(String input, String command) throws IllegalArgumentException, IndexOutOfBoundsException, NumberFormatException {
```
sevenseasofbri: Lines over 120 chars long should be wrapped to the next line

src/main/java/Bond/Bond.java
```diff
@@ -0,0 +1,325 @@
...
+            case DEADLINE:
+                addDeadline(input);
+                break;
+            case EVENT:
+                addEvent(input);
+                break;
+            case DELETE:
+                deleteTask(input);
+                break;
+            default:
```
sevenseasofbri: excellent job keeping everything at the same level of abstraction.

src/main/java/Bond/Bond.java
```diff
@@ -0,0 +1,325 @@
...
+
+    public static void executeCommand(String input) {
+        try {
+            String command = input.split(" ", 2)[0];
+            switch (command) {
+            case LIST:
+                printList();
+                break;
+            case MARK:
+            case UNMARK:
```
sevenseasofbri: > From java coding standard: The explicit` // Fallthrough` comment should be included whenever there is a case statement without a break statement.

src/main/java/Bond/Bond.java
```diff
@@ -0,0 +1,325 @@
...
+            if (dirsCreated) {
+                System.out.println("Directories created: " + parentDirectory.getAbsolutePath());
+            } else {
+                System.out.println("Failed to create directories!");
+            }
+        }
+        if (!taskFile.exists()) {
+            try {
+                if (taskFile.createNewFile()) {
+                    System.out.println("File created: " + taskFile.getAbsolutePath());
```
sevenseasofbri: This is around 3~ levels of indentation, consider refactoring the inner levels into their own functions

sevenseasofbri (commented):
Great job! Functions are very well written, I've given some minor suggestions.

PR #82, opened 2025-02-03
[kevinchangckc] iP

kevinchangckc:


text-ui-test/input.txt
noradazaperez, line 1: In the example you are missing the backslash /to

src/main/java/Viktor.java
```diff
@@ -187,3 +189,160 @@ public static void main(String[] args) {
...
+            // Exit program
+            if (input.equalsIgnoreCase("bye")) {
+                System.out.println("_________________________________________");
+                System.out.println("Bye. Hope to see you again soon.");
+                System.out.println("_________________________________________");
+                break;
+            }
+
+            // List all tasks
+            else if (input.equalsIgnoreCase("list")) {
```
noradazaperez, line 344: Maybe you can try to have smaller functions for each of the chatbot's options so that it doesn't get too long. I used a switch instead of if-else statements.

src/main/java/Viktor.java
```diff
@@ -0,0 +1,348 @@
...
+
+            // Unmark a task
+            else if (input.startsWith("unmark ")) {
+                try {
+                    int taskIndex = Integer.parseInt(input.split(" ")[1]) - 1;
+                    tasks.get(taskIndex).markAsNotDone();
+                    System.out.println("_________________________________________");
+                    System.out.println("OK, I've marked this task as not done yet:");
+                    System.out.println(tasks.get(taskIndex));
+                    System.out.println("_________________________________________");
```
yx-tzzz, line 425: Better to define the line separator as a constant or create a utility method for the line separator, for better maintainability

src/main/java/Viktor.java
```diff
@@ -0,0 +1,348 @@
...
+        this.from = from;
+        this.to = to;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (from: " + from + " to: " + to + ")";
+    }
+}
+
```
yx-tzzz: Maybe it would be better to separate each class into it's own file

src/main/java/Viktor.java
```diff
@@ -0,0 +1,348 @@
...
+            // Invalid command
+            else {
+                System.out.println("Sorry, I don't understand that command.");
+            }
+        }
+
+        scanner.close();
+    }
+}
```
yx-tzzz, line 463: Good use of variable and method names and camelCase

yx-tzzz, line 463: I like the comments before each block of code describing what the block does, and are indented relative to their position in the code

src/main/java/Viktor.java
```diff
@@ -0,0 +1,444 @@
+import java.util.Scanner;
+
```
Alfred-Goh02: Delete all your unused/commented codes

src/main/java/Viktor.java
```diff
@@ -0,0 +1,444 @@
...
+        scanner.close();
+    }
+}
+
+ */
+
+
+// level-3
+import java.util.ArrayList;
+
```
Alfred-Goh02: Separate your classes into their individual files

src/main/java/Viktor.java
```diff
@@ -0,0 +1,444 @@
...
+            try {
+                if (input.equalsIgnoreCase("bye")) {
+                    System.out.println("_________________________________________");
+                    System.out.println("Goodbye! Have a great day.");
+                    System.out.println("_________________________________________");
+                    break;
+                }
+
+                else if (input.equalsIgnoreCase("list")) {
```
Alfred-Goh02: Use K&R style brackets for else if, you can consider using switch instead

src/main/java/Viktor.java
```diff
@@ -0,0 +1,444 @@
...
+  */
+
+// Level-5
+class ViktorException extends Exception {
+    public ViktorException(String message) {
+        super(message);
+    }
+}
+
+public class Viktor {
```
Alfred-Goh02: Consider separate all the printing into a separate class

src/main/java/Viktor.java
```diff
@@ -0,0 +1,444 @@
...
+        System.out.println("_________________________________________");
+        System.out.println("Hi there. I'm Viktor.");
+        System.out.println("How can I help you?");
+        System.out.println("_________________________________________");
+
+        while (true) {
+            String input = scanner.nextLine();
+
+            try {
+                if (input.equalsIgnoreCase("bye")) {
```
Alfred-Goh02: For command strings like list, bye etc, avoid magic strings and declare it at the start of the file as a constant.

PR #83, opened 2025-02-03
[Neeraj Kumbar] iP

flyingapricot:


src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+    while (!(line.equals("bye"))) {
+      System.out.println(HORIZONTAL_LINE);
+      System.out.println("\t" + line);
+      System.out.println(HORIZONTAL_LINE);
+      line = in.nextLine();
+    }
+    goodbye();
+  }
+
+  public static void ListTasks() {
```
ojassurana: Should the method name ListTasks be changed to listTasks to follow Java’s camelCase convention? I noticed the same issue in several other places too.

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+  private static final String CHATBOT_NAME = "ORANGE";
+  private static final ArrayList<Task> tasks =
+      new ArrayList<Task>(); // Array List to store tasks entered by users
+
+  public static void greeting() {
+    System.out.println(HORIZONTAL_LINE);
+    System.out.println("\t" + "Hello! I'm " + CHATBOT_NAME);
+    System.out.println("\t" + "What can I do for you?");
+    System.out.println(HORIZONTAL_LINE);
+  }
```
ojassurana: Perhaps, the indentation should 4 spaces and not tab according to java conventions? I noticed the same issue in several other places too.

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Scanner;
+
+public class Orange {
+  private static final String HORIZONTAL_LINE = "\t" + "-".repeat(50);
+  private static final HashSet<String> commands = new HashSet<>(Arrays.asList("mark","unmark","list","todo","event","deadline"));
+
+  private static final String CHATBOT_NAME = "ORANGE";
```
ojassurana: I like how you named CHATBOT_NAME in uppercase, as it perfectly follows Java's naming convention for constants. 👍

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+  private static final ArrayList<Task> tasks =
+      new ArrayList<Task>(); // Array List to store tasks entered by users
+
+  public static void greeting() {
```
ojassurana: Perhaps, maybe the method greeting() could be renamed to printGreeting() to make it more action-oriented?

ojassurana (commented):
Overall looks good to me.
Just need some cosmetic changes to adhere to the java coding standards such as [Layout](https://se-education.org/guides/conventions/java/basic.html#layout) and [Naming](https://se-education.org/guides/conventions/java/basic.html#naming)

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Scanner;
+
+public class Orange {
+  private static final String HORIZONTAL_LINE = "\t" + "-".repeat(50);
+  private static final HashSet<String> commands = new HashSet<>(Arrays.asList("mark","unmark","list","todo","event","deadline"));
```
gitHST: Nice

src/main/java/Orange.java
gitHST: Very cool website to generate some logo art if you would like:
https://patorjk.com/software/taag/#p=display&f=Graffiti&t=ORANGE

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+    System.out.println(HORIZONTAL_LINE);
+    System.out.println("\t" + "Hello! I'm " + CHATBOT_NAME);
+    System.out.println("\t" + "What can I do for you?");
+    System.out.println(HORIZONTAL_LINE);
+  }
+
+  public static void goodbye() {
+    System.out.println(HORIZONTAL_LINE);
+    System.out.println("\t" + "Bye. Hope to see you again soon!");
+    System.out.println(HORIZONTAL_LINE);
```
gitHST: I notice there are many places where you print horizontal lines above and below some text. You could create a method `lineWrapPrint(String text)` that takes some text and prints it with horizontal lines above and below? in the long run would probably save on space since it saves 2 lines per usage.

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  public static void echo() {
+    String line;
+    Scanner in = new Scanner(System.in);
+    line = in.nextLine();
+    while (!(line.equals("bye"))) {
+      System.out.println(HORIZONTAL_LINE);
+      System.out.println("\t" + line);
+      System.out.println(HORIZONTAL_LINE);
```
gitHST: Example of place where `lineWrapPrint` could theoretically be used

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  public static void ListTasks() {
+    System.out.println(HORIZONTAL_LINE);
+    System.out.println("\tHere are the tasks in your list:");
+    int count = 1;
+    for (int i = 0; i < tasks.size(); i++) {
+      System.out.println("\t" + count + ". " + tasks.get(i).GetTaskWithCompletion());
+      count++;
+    }
```
gitHST:
```
String result = IntStream.range(0, tasks.size())
    .mapToObj(i -> "\t" + (i + 1) + ". " + tasks.get(i).GetTaskWithCompletion())
    .collect(Collectors.joining("\n"));

lineWrapPrint(result);
```
A bit of a stretch but would theoretically work haha

gitHST: Would need to be `lineWrapPrint(result + "\tHere are the tasks in your list:");`, My mistake

gitHST: other way round haha my bad

src/main/java/Orange.java
```diff
@@ -0,0 +1,178 @@
...
+    System.out.println(HORIZONTAL_LINE);
+  }
+
+  public static void AddTask(String task) {
+    System.out.println(HORIZONTAL_LINE);
+    System.out.println("\t" + "added: " + task);
```
gitHST: Nice usage of \t, i had never seen before so you learn something new everyday

gitHST (commented):
Very nice overall, just minor conciseness suggestions

src/main/java/orange/Orange.java
```diff
@@ -0,0 +1,332 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Scanner;
+import orange.task.Deadline;
+import orange.task.Events;
+import orange.task.Task;
+import orange.task.Todo;
+import java.io.IOException;
+import java.nio.file.*;
```
Paulifyer: Please avoid using this import type, import each package you need individually

src/main/java/orange/Orange.java
```diff
@@ -0,0 +1,332 @@
...
+import java.io.File;  // Import the File class
+import java.io.FileNotFoundException;  // Import this class to handle errors
+
+
+public class Orange {
+  private static final String HORIZONTAL_LINE = "\t" + "-".repeat(50);
+  private static final HashSet<String> commands = new HashSet<>(Arrays.asList("mark","unmark","list","todo","event","deadline","delete"));
```
Paulifyer: Consider using an enum here instead

src/main/java/orange/Orange.java
```diff
@@ -0,0 +1,332 @@
...
+      System.out.println("File exists, reading task file: " + TASKFILE);
+    } else {
+      // File doesn't exist, so create it
+      try {
+        Files.createFile(path);
+        System.out.println("Task File created at: " + TASKFILE);
+      } catch (IOException e) {
+        System.out.println("Error creating Task file: " + e.getMessage());
+      }
+    }
```
Paulifyer: try break up the file creation into another method

src/main/java/orange/Orange.java
```diff
@@ -0,0 +1,332 @@
...
+      Files.write(path, Arrays.asList(finalTask), StandardOpenOption.APPEND);
+      System.out.println("Data written to CSV successfully.");
+    } catch (IOException e) {
+      System.out.println("Error writing to CSV file: " + e.getMessage());
+    }
+
+  }
+
+  //Adding orange.task.Deadline orange.task.Task
+  public static void Deadline(String task,String dateAndTime,Boolean deadlineStatus) {
```
Paulifyer: the boolean name deadlineStatus does not follow the naming convention for booleans

src/main/java/orange/Orange.java
```diff
@@ -0,0 +1,332 @@
...
+        if(!commands.contains(commandWord)) throw new OrangeException(1);
+      }catch(OrangeException o) {
+        System.out.println(HORIZONTAL_LINE);
+        System.out.println("\t" + o.getMessage());
+        System.out.println(HORIZONTAL_LINE);
+      }
+
+    }
+    goodbye();
+  }
```
Paulifyer: maybe try to break down the code here into seperate individual methods.

PR #85, opened 2025-02-03
[xyanjun02]iP

xyanjun02:


src/main/java/BOB.java
```diff
@@ -1,51 +1,138 @@
...
+            }
+        }
+        scanner.close();
+    }
+
+    private static void printDivider() {
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void printGoodbyeMessage() {
```
randust, line 66: I like your way to refactor the code into static functions

randust (approved):

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "[" + getStatusIcon() + "] " + description;
+    }
+}
+
+public class BOB {
+    private static final int MAX_TASKS = 100;
```
lwenyi1: Good naming style for constant

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
+import java.util.Scanner;
+
+class Task {
```
lwenyi1: Consider moving `Task` class to a separate file to keep the main file cleaner

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
...
+        return isDone ? "X" : " ";
+    }
+
+    @Override
+    public String toString() {
+        return "[" + getStatusIcon() + "] " + description;
+    }
+}
+
+public class BOB {
```
lwenyi1: Consider changing the class name to `Bob` instead. Can still leave the bots actual name as BOT when printing etc but this would help people immediately recognize `Bob` as a class instead of a constant in the code

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
...
+    private static int taskCount = 0;
+    private static final Scanner scanner = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        printDivider();
+        System.out.println(" Hello! I'm BOB");
+        System.out.println(" What can I do for you? 😀");
+        printDivider();
+
+        while (true) {
```
lwenyi1: Consider adding a variable along the lines of `isRunning` to make it clearer what terminates the program instead of using a break within the loop

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
...
+            } else {
+                System.out.println(" Invalid task number!");
+            }
+        } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
+            System.out.println(" Please enter a valid task number.");
+        }
+    }
+
+    private static void handleUnmark(String[] parts) {
+        try {
```
lwenyi1: Good use of try catch to handle bogus user input

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
...
+            int taskNumber = Integer.parseInt(parts[1]) - 1;
+            if (isValidTaskNumber(taskNumber)) {
+                tasks[taskNumber].unmarkAsDone();
+                printDivider();
+                System.out.println(" OK, I've marked this task as not done yet:");
+                System.out.println("   " + tasks[taskNumber]);
+                printDivider();
+            } else {
+                System.out.println(" Invalid task number!");
+            }
```
lwenyi1: Clear but can be slightly improved by considering the if statement:
```
if (!isValidTaskNumber(taskNumber)) {
    System.out.println(" Invalid task number!");
    return;
}
// <code in original if block>
```
so less code is nested. Similarly for the previous try catch

src/main/java/BOB.java
```diff
@@ -0,0 +1,138 @@
...
+        printDivider();
+        System.out.println(" Hello! I'm BOB");
+        System.out.println(" What can I do for you? 😀");
+        printDivider();
+
+        while (true) {
+            String input = scanner.nextLine().trim();
+            String[] parts = input.split(" ", 2);
+
+            if (input.equalsIgnoreCase("bye")) {
```
lwenyi1: Use of functions to print messages or handle inputs makes code simple and readable here :-)

lwenyi1 (commented):
Very clean and readable code.

Can consider using separate files for the other classes to keep the main Bob file clean and for more abstraction.

src/main/java/Bob.java
```diff
@@ -0,0 +1,171 @@
...
+    private static int taskCount = 0;
+    private static final Scanner scanner = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        printDivider();
+        System.out.println(" Hello! I'm BOB");
+        System.out.println(" What can I do for you? 😀");
+        printDivider();
+
+        while (true) {
```
bljhty: you may want to have a boolean variable as a flag instead of using while(true) as while(true) is generally not advisable

src/main/java/Bob.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private static void printDivider() {
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void printGoodbyeMessage() {
+        printDivider();
+        System.out.println(" Bye! Hope to see you again soon! 😀");
+        System.out.println(" - BOB");
```
bljhty: in general, magic strings is not recommended and you might want to consider putting these magic strings as constants instead.

PR #87, opened 2025-02-03
[Roma637] iP

Roma637:
Level 1, 2, 3


src/main/java/Garfield.java
```diff
@@ -0,0 +1,73 @@
+import java.util.Scanner;
+
+public class Garfield {
+
+    public static void Respond(String sentence) {
+        System.out.println("------------------");
```
ShengJie13245, line 12: You should use a constant like SEPERATOR = "---------"

src/main/java/Garfield.java
```diff
@@ -0,0 +1,73 @@
...
+    public static void main(String[] args) {
+
+        Logo();
+        Greet();
+
+        Scanner input = new Scanner(System.in);
+        String line = input.nextLine();
+        boolean exit = false;
+
+        List l = new List(new String[]{});
```
ShengJie13245: A better name than l should be used

src/main/java/Garfield.java
```diff
@@ -0,0 +1,73 @@
...
+        Respond("Fine, I'm leaving to find more lasagna.");
+    }
+
+    public static void main(String[] args) {
+
+        Logo();
+        Greet();
+
+        Scanner input = new Scanner(System.in);
+        String line = input.nextLine();
```
ShengJie13245: line is not a intuitive name for input

src/main/java/Garfield.java
```diff
@@ -0,0 +1,73 @@
+import java.util.Scanner;
+
+public class Garfield {
+
+    public static void Respond(String sentence) {
```
ShengJie13245, line 11: Maybe a method name like printResponse would be better, the first letter of a method name should be lower case

src/main/java/List.java
```diff
@@ -0,0 +1,80 @@
+public class List {
+//    private String[] taskList ;
+//    private boolean[] completed ;
+
+    private Task[] taskList ;
+    private int size ;
```
kmnyn: Very clear task management.

src/main/java/List.java
```diff
@@ -0,0 +1,80 @@
...
+            whole += ":[" ;
+            if (this.taskList[i].isDone()) {
+                whole += "X";
+            } else {
+                whole += " ";
+            }
+            whole += "] " + this.taskList[i].getTaskName() + "\n";
+
+//            System.out.println(whole);
+        }
```
kmnyn: Readable and well-structured task display.

src/main/java/List.java
```diff
@@ -0,0 +1,80 @@
...
+        return("added task: " + task);
+    }
+
+    public String markTask(int taskIndex) {
+        if (taskIndex < size) {
+            taskList[taskIndex].setDone(true);
+            return "I've marked this as done: " + taskList[taskIndex].getTaskName();
+        }
+        return "ERROR. Index not in list";
+    }
```
kmnyn: Would it be better to also check for negative values using taskIndex >= 0?

src/main/java/List.java
```diff
@@ -0,0 +1,80 @@
+public class List {
+//    private String[] taskList ;
+//    private boolean[] completed ;
```
kmnyn: Would it be better to remove the unused complete[ ]?

kmnyn (commented):
Overall I think your code is well-structured and functional. Nice job done!

src/main/java/Garfield.java
```diff
@@ -0,0 +1,73 @@
...
+        System.out.println("------------------");
+        System.out.print("\t");
+        System.out.println(sentence);
+        System.out.println("------------------");
+    }
+
+    public static void Logo() {
```
xiaohan28, line 18: all methods should be camelCase, so first letter should be lowercase. Methods should ideally be verbs, like printLogo.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    private String taskName;
```
xiaohan28: If taskName is not suppose to change once created can consider making taskName final.

src/main/java/Task.java
xiaohan28, line 1: Good use of constructors and encapsulation with getters and setters.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public boolean isDone() {
+        return done;
+    }
+
+    public void setDone(boolean done) {
+        this.done = done;
+    }
+
```
xiaohan28, line 27: Might want to add methods like markDone() and markUndone() for better readability instead of directly using setDone(true/false) in List.

src/main/java/Garfield.java
```diff
@@ -0,0 +1,235 @@
...
+
+    public static void Greet() {
+        Respond("Oh. It's you. Hello. What do you want?");
+    }
+
+    public static void Exit() {
+        Respond("Fine, I'm leaving to find more lasagna.");
+    }
+
+public static List ReadFromFile(String filePath) throws TaskException {
```
aditichadha1310: Avoid lengthy methods. try to keep each method length to about 30 lines of code.

src/main/java/Garfield.java
```diff
@@ -0,0 +1,235 @@
...
+            ArrayList<Task> tasks = l.getTaskList();
+            int size = l.getSize();
+
+            for (int i = 0; i < size; i++) {
+                Task task = tasks.get(i);
+                StringBuilder line = new StringBuilder();
+
+                // Add type
+                if (task instanceof Todo) {
+                    line.append("T | ");
```
aditichadha1310, line 110: Avoid magic strings and magic numbers in the code.

PR #90, opened 2025-02-03
[ashertan256] iP

Ashertan256:


src/main/java/Alex.java
```diff
@@ -3,7 +3,7 @@
 public class Alex {
     public static void main(String[] args) {
         Scanner scanner = new Scanner(System.in);
-        String[] tasks = new String[100];
+        Task[] tasks = new Task[100];
```
edwardrl101: Perhaps extracting 100 out as a constant might work here to prevent it from being a magic number?

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+                    System.out.println("    ____________________________________________________________");
+                    for (int i = 0; i < taskCount; i++) {
+                        System.out.println("    "+(i + 1) + ".["+tasks[i].getStatusIcon()+"] " + tasks[i].description);
+                    }
+                    System.out.println("    ____________________________________________________________");
+                }
+            }else if (input.contains("unmark")) {
+                //unmarked placed before to prevent triggering on "mark"
+                System.out.println("    ____________________________________________________________");
+                int target = Integer.parseInt(input.substring(input.indexOf(" ")+ 1, input.length()));
```
edwardrl101: Perhaps a more informative and intuitive variable name for target?

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
+import java.util.Scanner;
+
+public class Alex {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
```
edwardrl101: I like how you followed the camelCase naming convention for variables!

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
edwardrl101, line 6: I like how you followed the naming convention for Boolean variables

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    
+    public void setMark() {
+        this.isDone = true;
+    }
+
+    public void setUnmark() {
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
```
edwardrl101, line 23: The naming for the getter method is good

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+    
+    public void setMark() {
+        this.isDone = true;
+    }
+
+    public void setUnmark() {
```
edwardrl101, line 19: I like how you followed the naming convention for setter methods

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+    
+    public void setMark() {
```
edwardrl101, line 15: Good job following the camelCase naming convention for methods

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+
+        System.out.println("Hello! I'm Alex");
+        System.out.println("What can I do for you?");
+        System.out.println();
+
+        String input;
+        do {
+            input = scanner.nextLine();
+
+            if (input.equalsIgnoreCase("list")) {
```
HTY2003: To avoid arrowhead style code, you can consider handling unusual input cases by putting the input handling code for a given command into its own function. After this, you can catch unusual cases and return/end the function early.

E.g. brief pseudocode for function contents

if (unusualCase)
    return;

if (unusualCase2)
    return;

continueHandlingInput();

This will make the code less confusing (although this is a great attempt)

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+                if (taskCount == 0) {
+                    System.out.println("No tasks");
+                } else {
+                    System.out.println("    ____________________________________________________________");
+                    for (int i = 0; i < taskCount; i++) {
+                        System.out.println("    "+(i + 1) + ".["+tasks[i].getStatusIcon()+"] " + tasks[i].description);
```
HTY2003: Perhaps to improve readability and shorten the long line of code, this could be wrapped in its own function/class

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+                //unmarked placed before to prevent triggering on "mark"
+                System.out.println("    ____________________________________________________________");
+                int target = Integer.parseInt(input.substring(input.indexOf(" ")+ 1, input.length()));
+                tasks[target - 1].setUnmark();
+                System.out.println("    Unmarked "+tasks[target-1].description);
+                System.out.println("    ____________________________________________________________");
+            } 
+            else if (input.contains("mark")) {
+                System.out.println("    ____________________________________________________________");
+                int target = Integer.parseInt(input.substring(input.indexOf(" ")+ 1, input.length()));
```
HTY2003: Try to split this line up, to avoid an overly complicated expression

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
+import java.util.Scanner;
+
+public class Alex {
```
HTY2003, line 3: This method is over 30 lines long. Consider splitting it into parts to improve readability

felfelyuen: I agree with this, as splitting this method into different methods, such as handleGoodbye and handleAdd for example, could be useful as you would be able to use these in other methods as the iP becomes more complicated

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+                System.out.println("    ____________________________________________________________");
+            } 
+            else if (input.contains("mark")) {
+                System.out.println("    ____________________________________________________________");
+                int target = Integer.parseInt(input.substring(input.indexOf(" ")+ 1, input.length()));
+                tasks[target - 1].setMark();
+                System.out.println("    Marked "+tasks[target-1].description);
+                System.out.println("    ____________________________________________________________");
+            } 
+            else if (!input.equalsIgnoreCase("bye")) {
```
felfelyuen: You could consider implementing the if statement as 
"if (!input.equalsIgnoreCase("bye") && (taskCount < 100) )"
to avoid arrowhead, or implement a switch function instead as well.

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+
+                    taskCount++;
+                    System.out.println("    ____________________________________________________________");
+                    System.out.println("    added: " + input);
+                    System.out.println("    ____________________________________________________________");
+                }
+                
+            }
+        } while (!input.equalsIgnoreCase("bye"));
+        System.out.println("    ____________________________________________________________");
```
felfelyuen: Small issue, but you could consider implementing a method/class to print the line, as you may need to print lines over many classes. This can improve readability too.

src/main/java/Task.java
felfelyuen: I agree, I also like how verbs are used in these methods, and nouns for the variables, the code is more readable because of this.

felfelyuen (commented):
Good in following the coding standards generally, though I think the code could be made more readable by splitting the methods.

src/main/java/Alex.java
```diff
@@ -0,0 +1,60 @@
...
+                if (taskCount == 0) {
+                    System.out.println("No tasks");
+                } else {
+                    System.out.println("    ____________________________________________________________");
+                    for (int i = 0; i < taskCount; i++) {
+                        System.out.println("    "+(i + 1) + ".["+tasks[i].getStatusIcon()+"] " + tasks[i].description);
+                    }
+                    System.out.println("    ____________________________________________________________");
+                }
+            }else if (input.contains("unmark")) {
```
Lyam-T: should here has a space between } and else if?

src/main/java/Alex.java
Lyam-T: maybe the line is a bit to long, perhaps can move to different line

src/main/java/Alex.java
```diff
@@ -0,0 +1,281 @@
...
+
+        Scanner scanner = new Scanner(System.in);
+
+        while (true) {
+            String input = scanner.nextLine().trim();
+
+            // Check for "bye" first
+            if (input.equalsIgnoreCase("bye")) {
+                System.out.println("    ____________________________________________________________");
+                System.out.println("    Ba...Bak...Baka Baka!! I didn't want to talk to you anyway....");
```
aditichadha1310: Please keep the tone of the bot very formal. Avoid usage of non-english words.

src/main/java/Alex.java
```diff
@@ -0,0 +1,281 @@
...
+                    parts[i] = parts[i].trim();
+                }
+                if (parts.length < 3) {
+                    continue;
+                }
+                String taskType = parts[0];
+                boolean isDone = Boolean.parseBoolean(parts[1]);
+                String description = parts[2];
+                Task task = null;
+                if (taskType.equals("T")) {
```
aditichadha1310: Avoid using Magic Strings

src/main/java/Alex.java
```diff
@@ -0,0 +1,281 @@
...
+        try {
+            int index = Integer.parseInt(input.substring(6).trim());
+            if (index < 1 || index > tasks.size()) {
+                throw new IndexOutOfBoundsException();
+            }
+            Task deletedTask = tasks.get(index - 1);
+            deletedTask.delete();
+            tasks.remove(index - 1);
+            System.out.println("    ____________________________________________________________");
+            System.out.println("     H-Hey! Don’t get the wrong idea! I-it's not like I want to delete this task for you or anything… Hmph! But don’t think I’ll do this every time, okay?! Jeez…");
```
aditichadha1310: Please limit the number of characters to be max 120 per line of code.

src/main/java/Alex.java
```diff
@@ -0,0 +1,281 @@
...
+     */
+    private static void handleUnmark(String input) {
+        try {
+            int index = Integer.parseInt(input.substring(6).trim());
+            if (index < 1 || index > tasks.size()) {
+                throw new IndexOutOfBoundsException();
+            }
+            tasks.get(index - 1).setUnmark();
+            System.out.println("    ____________________________________________________________");
+            System.out.println("     Gambatte, I've marked this task as not done yet:");
```
aditichadha1310: Avoid usage of non-english words as this application caters to international audience

PR #95, opened 2025-02-03
[ecxm02] iP

ecxm02:


src/main/java/ParseText.java
```diff
@@ -21,27 +25,22 @@ private void textCleanup() {
         }
 
         else if (textInput.startsWith("deadline")) {
```
clj55: 'else if' and 'else' should be on the same line as the previous }

src/main/java/ParseText.java
```diff
@@ -21,27 +25,22 @@ private void textCleanup() {
...
 
         else if (textInput.startsWith("event")) {
-            int spaceIndex = textInput.indexOf(" ");
             inputCommand = textInput.substring(0, spaceIndex);
-            int slashIndex = textInput.indexOf("/", spaceIndex + 1);
-            taskName = textInput.substring(spaceIndex + 1, slashIndex - 1);
-            int slashIndex2 = textInput.indexOf("/", slashIndex + 1);
-            from = textInput.substring(slashIndex + 6, slashIndex2 - 1);
-            to = textInput.substring(slashIndex2 + 4);
+            taskName = textInput.substring(spaceIndex + Constants.EMPTY_PAD, slashIndex - Constants.EMPTY_PAD);
```
clj55: I like how you shorten it, makes it more readable :D

src/main/java/ParseText.java
```diff
@@ -21,27 +25,22 @@ private void textCleanup() {
...
         }
 
         else if (textInput.startsWith("todo")) {
-            int spaceIndex = textInput.indexOf(" ");
             inputCommand = textInput.substring(0, spaceIndex);
-            taskName = textInput.substring(spaceIndex + 1);
+            taskName = textInput.substring(spaceIndex + Constants.EMPTY_PAD);
         }
 
         else {
```
clj55: here too, else should be on the same line as the previous }

src/main/java/Juke.java
```diff
@@ -3,7 +3,8 @@
 public class Juke {
     private static boolean isCompleted = false; //a master flag so that main knows when to end the program
 
-    public static void endProgram() { //sets the isCompleted flag to true so that the loop ends, also prints end message
+    //sets the isCompleted flag to true so that the loop ends, also prints end message
```
clj55: I think comments about the method (indented on same level as endProgram like how you have done) should describe what the method does overall. If you want to explain how the code works maybe put it inside the the method (under end Program and indented to match isCompleted

clj55 (commented):
Overall clean code yay :D 
Side note: you can use the string.split function instead of getting the index then doing substring

src/main/java/Juke.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public static void main(String[] args) {
+        System.out.println(Constants.INTRO_MESSAGE);
+        Scanner input = new Scanner(System.in);
+        String text;
+
+        while (!isCompleted) {
+            text = input.nextLine();
+            ReplyText current = new ReplyText(text);
```
dylankhoo: Seeing that you are not using the "current" variable elsewhere, rather than creating new instances of ReplyText, could it be better to just call a method from ReplyText here?

src/main/java/ReplyText.java
```diff
@@ -0,0 +1,46 @@
+public class ReplyText {
+    private static TaskManager taskList = new TaskManager();
+    private String textInput;
+
+
+    public ReplyText(String textInput) {
+        this.textInput = textInput;
+        ParseText cleaner = new ParseText(textInput);
```
dylankhoo: A more intuitive variable name here could perhaps be "cleanedText"

src/main/java/ParseText.java
```diff
@@ -0,0 +1,74 @@
...
+
+        else if (textInput.startsWith("todo")) {
+            inputCommand = textInput.substring(0, spaceIndex);
+            taskName = textInput.substring(spaceIndex + Constants.EMPTY_PAD);
+        }
+
+        else {
+            inputCommand = textInput;
+        }
+    }
```
dylankhoo: This is quite a long method, perhaps you could consider breaking it into smaller methods? You could have individual methods (e,.g. cleanUpMark) for each command and call them accordingly based on textInput.

src/main/java/ReplyText.java
```diff
@@ -0,0 +1,46 @@
+public class ReplyText {
+    private static TaskManager taskList = new TaskManager();
```
dylankhoo: Rather than creating the TaskManager instance here, could it be perhaps more intuitive to create the instance in Juke instead?

dylankhoo (commented):
Overall I like how you extracted your code into different classes and kept Juke clean.

However, I do find that in some places, it may be more appropriate to call static methods rather than constructing instances of new classes. 

I have left some comments on where I felt that this was the case.

src/main/java/ReplyText.java
ecxm02: Ok! Will amend

src/main/java/ParseText.java
ecxm02: Alright

src/main/java/ReplyText.java
ecxm02: Okays

src/main/java/Juke.java
ecxm02: Yea I will change some classes into static and call their methods instead

src/main/java/Juke.java
ecxm02: Okays

src/main/java/ParseText.java
ecxm02: Okay!

src/main/java/juke/input/InputHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    private InputHandler() {
+    }
+
+    public static void readText(TaskManager taskManager, String textInput) {
+        Parser.cleanText(textInput);
+        respondToText(taskManager, Parser.getInputCommand(), Parser.getTaskName(), Parser.getTaskIndex(),
+                Parser.getDeadline(), Parser.getFrom(), Parser.getTo());
+    }
+
+    //uses a switch to execute appropriate methods for the input
```
sevenseasofbri: Implementation details like this can be left out of the comments. This may not be providing any useful input to the reader.

src/main/java/juke/input/InputHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void respondToText(TaskManager taskManager, String inputCommand, String taskName, int taskIndex,
+                                     String deadline, String from, String to) {
+
+        switch (inputCommand) {
```
sevenseasofbri, line 42: Good job keeping a consistent level of abstraction.

src/main/java/juke/input/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+    private static String inputCommand = "invalid";
+    private static String taskName;
+    private static String taskNumber;
+    private static int taskIndex = -1;
+    private static int spaceIndex = -1;
+    private static int slashIndex = -1;
+    private static int slashIndex2 = -1;
+    private static String deadline;
+    private static String from;
+    private static String to;
```
sevenseasofbri: If any of these are not going to change in the duration of the program, you can change them to constants `static final` in `SCREAMING_SNAKE_CASE`

src/main/java/juke/input/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private Parser() {
+    }
+
+    public static void cleanText(String text) {
```
sevenseasofbri, line 34: This function is slightly >30 LOC. Consider chunking it into multiple methods.

src/main/java/juke/input/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+                cleanUpEvent();
+            } catch (Exception e) {
+                inputCommand = "invalid";
+            }
+        } else if (textInput.startsWith("delete")) {
+            try {
+                cleanUpDelete();
+            } catch (Exception e) {
+                inputCommand = "invalid";
+            }
```
sevenseasofbri: You can refactor these try catch blocks into their own methods.

src/main/java/juke/persistence/FileSaver.java
```diff
@@ -0,0 +1,60 @@
...
+            case "D":
+                Deadline deadlineTask = (Deadline) task;
+                writer.write("D" + "|" + task.isDone() + "|" + deadlineTask.getTaskName() + "|"
+                        + deadlineTask.getDeadline() + System.lineSeparator());
+                break;
+            case "E":
+                Event eventTask = (Event) task;
+                writer.write("E" + "|" + task.isDone() + "|" + eventTask.getTaskName() + "|"
+                        + eventTask.getFrom() + "|" + eventTask.getTo() + System.lineSeparator());
+                break;
```
sevenseasofbri: Consider extracting these complex expressions into thir own methods.

src/main/java/juke/task/TaskManager.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.ArrayList;
+
+public class TaskManager {
+    private ArrayList<Task> taskList = new ArrayList<>();
+
+    public TaskManager() {
+    }
+
+    public void listTask() { //prints out the tasks and marks them appropriately
```
sevenseasofbri: Consider using JavaDoc format to describe methods instead.

sevenseasofbri (commented):
Great job so far! Some segments of the code have deep nesting (>=3 levels of indentation, e.g.. `readFromFile` from FileLoader) so you might want to try and look resloving that.

PR #101, opened 2025-02-04
[Li-JunXian] iP

Li-JunXian:


src/main/java/Lintz.java
```diff
@@ -26,19 +26,32 @@ public static void echo(String input) {
     }
```
jyukuan, line 25: I do not like the echo function since it  can just be replaced by system.out.println(input).

src/main/java/Lintz.java
```diff
@@ -26,19 +26,32 @@ public static void echo(String input) {
...
     public static void taskListManager(String input) {
-        if (!input.equalsIgnoreCase("list")){
-            taskList[taskCount] = input;
+        if (input.equalsIgnoreCase("list")) {
             System.out.println("\t" + horiLine);
-            System.out.print("\t" + "added: ");
-            echo(input);
+            System.out.println("\t" + "Here are the tasks in your list:");
+            for (int i = 0; i < taskCount; i++) {
+
```
jyukuan, line 28: It might be an additional blank line.

src/main/java/Task.java
```diff
@@ -0,0 +1,28 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
jyukuan: I like the way you present the condition checking! It is so easy to read.

src/main/java/Task.java
```diff
@@ -0,0 +1,28 @@
+
+public class Task {
```
jyukuan, line 4: May I ask it is any reason to put class task in a different file?

src/main/java/Lintz.java
```diff
@@ -8,7 +8,7 @@ public class Lintz {
             + "    | |___  | | |  _  |  | |__  /  /       \n"
             + "    |_____| |_| |_| |_|  |___/ /____|      \n";
     private static String horiLine = "\u2500".repeat(100);
-    private static String[] taskList = new String[100];
+    private static Task[] taskList = new Task[100];
     private static int taskCount = 0;
 
     public static void greet(){
```
jyukuan: Any reason to leave that function here since the function is not used

jyukuan (commented):
Looks great! Pretty clean and easy to understand!

src/main/java/Task.java
itsjoelha, line 4: I think this class file was nicely implemented.

src/main/java/Lintz.java
```diff
@@ -8,7 +8,7 @@ public class Lintz {
             + "    | |___  | | |  _  |  | |__  /  /       \n"
             + "    |_____| |_| |_| |_|  |___/ /____|      \n";
     private static String horiLine = "\u2500".repeat(100);
```
itsjoelha: Perhaps you could declare this as a final as it is likely not going to change.

src/main/java/Lintz.java
```diff
@@ -26,19 +26,32 @@ public static void echo(String input) {
...
+            for (int i = 0; i < taskCount; i++) {
+
+                System.out.println("\t" + (i + 1) + ".[" + taskList[i].getStatusIcon() + "] " + taskList[i].description);
+            }
             System.out.println("\t" + horiLine);
-            taskCount += 1;
+
+        } else if (input.toLowerCase().startsWith("mark")) {
+            System.out.println("\t" + horiLine);
+            taskList[Integer.parseInt(input.substring(5)) - 1].markAsDone();
```
itsjoelha: Perhaps you could check if the index accessed would be out of bounds?

src/main/java/Lintz.java
```diff
@@ -26,19 +26,32 @@ public static void echo(String input) {
...
+            System.out.println("\t" + horiLine);
+
+        } else if (input.toLowerCase().startsWith("unmark")) {
+            System.out.println("\t" + horiLine);
```
itsjoelha: Instead maybe you could perhaps add the "\t" to the constant?

itsjoelha (commented):
I think it was general well written.

src/main/java/Lintz.java
jyukuan: Got it, Thanks!

src/main/java/Lintz.java
jyukuan: Got it, Thanks! I will add another function to check it.

src/main/java/Lintz.java
jyukuan: Got it, thanks!

src/main/java/FileWrite.java
```diff
@@ -0,0 +1,67 @@
+import java.io.*;
```
wenxin-c: Please import libraries explicitly instead of using wildcard.

src/main/java/Lintz.java
```diff
@@ -0,0 +1,123 @@
...
+        System.out.println("\t " + "You have now " + (taskCount + 1) + " tasks Sir/Ma'am." + "\n");
+        System.out.println(horiLine + "\n");
+    }
+
+    public static void taskManager(String input) throws IllegalInputException {
+        String firstWord = input.split(" ")[0].toUpperCase();
+        String[] sentenceComponents = input.split("/");
+
+        try {
+            switch (firstWord) {
```
wenxin-c: You might consider abstracting each ```clause``` statement into a separate method so that the code will be neater and more organised.

wenxin-c (commented):
Well done for your IP and good use of OOP. Just minor suggestions for your consideration .

PR #107, opened 2025-02-04
[yiyang74] iP

yiyang74:


src/main/java/Beru.java
```diff
@@ -10,8 +11,7 @@ private static void printList(Task[] tasks) {
         System.out.print(LINE_SEPARATOR);
         System.out.println("Here are the tasks in your list: ");
         for (int i = 0; i < Task.getTaskCount(); i++) {
-            System.out.println((i + 1) +
-                    ".[" + tasks[i].getStatusIcon() + "] " + tasks[i].getDescription());
+            System.out.println((i + 1) + "." + tasks[i].toString());
```
vegetablestabber: `tasks[i].toString()` can just be `tasks[i]` when concatenating it to a String. It is, of course, more explicit to the reader but the reader can also figure out that the String representation is derived implicity from a `Task` instead.

src/main/java/Beru.java
```diff
@@ -57,6 +57,29 @@ private static void printAlreadyUnmarked() {
                 "Entry is already unmarked!\n" + LINE_SEPARATOR);
     }
 
+    private static void printTaskEntry(Task[] tasks) {
+        System.out.print(LINE_SEPARATOR + "Got it! I've added this task:\n  " +
+                tasks[Task.getTaskCount() - 1] + "\nNow you have " +
+                Task.getTaskCount() + " tasks in the list.\n" + LINE_SEPARATOR);
+    }
+
+    private static void printInvalidDeadline() {
```
vegetablestabber: I like how you have multiple methods to handle different types of invalid input, but maybe you could also consider throwing an `Exception` and handling it in the function where it is called?

This would allow you to disconnect printing from the error message and delegates the responsibility of handling the error to the method (e.g. `handleDeadlineCommand`) where the method that is throwing the `Exception` (e.g. `printInvalidDeadline`).

I noticed that you only handle errors by printing so shifting the responsibility to the parent method is something else you might find helpful as you can avoid confusion between using `print` or `println` for every error printing method, but I also understand the straightforwardness of your solution.

src/main/java/Beru.java
```diff
@@ -91,6 +114,61 @@ private static void handleListCommand(Task[] tasks) {
...
+            return;
+        }
+        String description = combineWordsToSentence(words, 1, indexOfBy);
+        String by = combineWordsToSentence(words, indexOfBy + 1, words.length);
+        tasks[Task.getTaskCount()] = new Deadline(by, description);
+        printTaskEntry(tasks);
+    }
+
+    private static void handleEventCommand(Task[] tasks, String[] words) {
+        int indexOfFrom = -1;
```
vegetablestabber: Maybe you could consider creating a new function for obtaining parameters from a given input? This might be useful for any future commands which require parameters and could also be used to handle missing parameters, thus reducing the size of methods like `handleDeadlineCommand` and delegating these responsibilities away into other modular methods.

src/main/java/Beru.java
```diff
@@ -91,6 +114,61 @@ private static void handleListCommand(Task[] tasks) {
...
+        tasks[Task.getTaskCount()] = new Event(from, to, description);
+        printTaskEntry(tasks);
+    }
+
+    private static String combineWordsToSentence(String[] words, int start, int end) {
+        words = Arrays.copyOfRange(words, start, end);
+        return String.join(" ", words);
+    }
+
     private static void parseUserSentence(String sentence, Task[] tasks) {
```
vegetablestabber: I like the simplicity of this function. A simple `switch` statement is clear and readable to me as a reviewer.

vegetablestabber (commented):
Looks great! Liked the modularity of your code and suggested some changes in that regard.

src/main/java/Beru.java
```diff
@@ -57,6 +57,29 @@ private static void printAlreadyUnmarked() {
...
+                tasks[Task.getTaskCount() - 1] + "\nNow you have " +
+                Task.getTaskCount() + " tasks in the list.\n" + LINE_SEPARATOR);
+    }
+
+    private static void printInvalidDeadline() {
+        System.out.print(LINE_SEPARATOR +
+                "Invalid entry for Deadline!\nPlease include \"/by\"!\n" + LINE_SEPARATOR);
+    }
+
+    private static void printMissingEventFields() {
```
OmJRamdas: Good Naming conventions, its easy to read

src/main/java/Beru.java
```diff
@@ -104,9 +182,14 @@ private static void parseUserSentence(String sentence, Task[] tasks) {
...
+            handleTodoCommand(tasks, words);
+            break;
+        case "deadline":
+            handleDeadlineCommand(tasks, words);
+            break;
+        case "event":
+            handleEventCommand(tasks, words);
             break;
         }
     }
```
OmJRamdas: Good job in avoiding nesting

src/main/java/Beru.java
```diff
@@ -104,9 +182,14 @@ private static void parseUserSentence(String sentence, Task[] tasks) {
         case "unmark":
             handleUnmarkedCommand(tasks, words);
             break;
-        default:
```
OmJRamdas: Could possibly abstract this from Beru to Task class

src/main/java/Beru.java
```diff
@@ -28,7 +28,7 @@ private static void printTaskStatus(Task t) {
         } else {
             System.out.println("OK, I've marked this task as not done yet:");
         }
-        System.out.println("  [" + t.getStatusIcon() + "] " + t.getDescription());
+        System.out.println("  " + t);
         System.out.print(LINE_SEPARATOR);
```
OmJRamdas: Create a function for printing LINE_SEPERATOR

src/main/java/Beru.java
yiyang74: Thanks for pointing it out! Forgot about this one 😅

src/main/java/Beru.java
yiyang74: oh good idea!

src/main/java/Beru.java
yiyang74: Do you mean handle the incorrect input format in Task?

src/main/java/Beru.java
yiyang74: oh is that better? I'm not sure

src/main/java/uncleroger/ui/UncleRoger.java
```diff
@@ -0,0 +1,300 @@
+package uncleroger.ui;
+
+import uncleroger.exception.*;
+import uncleroger.task.*;
+import java.util.Arrays;
+import java.util.Scanner;
+
```
Alfred-Goh02: Consider separating command handling methods into another class file, UI methods can also be separated into another class, event though this file is packaged as UI

src/main/java/uncleroger/ui/UncleRoger.java
```diff
@@ -0,0 +1,300 @@
...
+        words = Arrays.copyOfRange(words, start, end);
+        return String.join(" ", words);
+    }
+
+    private static void parseUserSentence(String sentence, Task[] tasks)
+            throws InvalidCommandException {
+        String[] words = sentence.split(" ");
+        String command = words[0];
+        switch (command.toLowerCase()) {
+        case "list":
```
Alfred-Goh02: Command strings such as list, mark etc should be declared as constants, avoid magic literals.

src/main/java/uncleroger/task/Deadline.java
```diff
@@ -0,0 +1,15 @@
+package uncleroger.task;
+
```
Alfred-Goh02, line 2: Good job, all your task classes are very concise

PR #116, opened 2025-02-04
[Giggs] iP

NotGiggs:


src/main/java/Bartholomew.java
```diff
@@ -0,0 +1,78 @@
...
+                        System.out.println("   " + tasks[index]);
+                        System.out.println("____________________________________________________________");
+                    } else {
+                        tasks[index].markAsNotDone();
+                        System.out.println("____________________________________________________________");
+                        System.out.println(" Absolutely, master! I have marked this task as incomplete.");
+                        System.out.println("   " + tasks[index]);
+                        System.out.println("____________________________________________________________");
+                    }
+                } catch (NumberFormatException e) {
```
trunghjieu2506: good use of exception

src/main/java/Bartholomew.java
kevinchangckc, line 1: I appreciate how you have empty lines between different sections to make the code readable

kevinchangckc, line 1: Nice customization of the chatbot's personality. I like what you did there

src/main/java/Task.java
kevinchangckc, line 1: I appreciate the comments next to some lines of code you have. Helps illustrate the purpose well

kevinchangckc, line 1: very clean way of structuring the methods and naming them in a clear manner

src/main/java/Bartholomew.java
```diff
@@ -0,0 +1,78 @@
+import java.util.Objects;
+import java.util.Scanner;
+
```
trunghjieu2506, line 2: Great code overall. You might consider add Javadoc comments in the code for better readability when the iP code size gets larger.

src/main/java/Bartholomew.java
```diff
@@ -0,0 +1,78 @@
...
+            String[] parts = input.split(" ", 2);
+            String command = parts[0];
+
+            if (command.equalsIgnoreCase("bye")) {
+                System.out.println("Goodbye Master! I bid you a jolly rest of your day!");
+                break;
+            }
+
+            else if (command.equalsIgnoreCase("list")) {
+                System.out.println("____________________________________________________________");
```
trunghjieu2506: It might be a good idea to create a method to print these repetitive seperators. 

Something along the line of 

private static void printSeparator() {
}

src/main/java/Bartholomew.java
```diff
@@ -0,0 +1,78 @@
...
+                System.out.println(" Here are the tasks in your list, master:");
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println((i + 1) + "." + tasks[i]);
+                }
+                System.out.println("____________________________________________________________");
+            }
+
+            else if (command.equalsIgnoreCase("mark") || command.equalsIgnoreCase("unmark")) {
+                if (parts.length < 2) {
+                    System.out.println("Kindly specify a task number, master.");
```
trunghjieu2506: It might be a good idea to provide the input format with this message. 

Something along the line of "Kindly specify a task number according to this format: mark <number>, master."

src/main/java/Bartholomew.java
```diff
@@ -0,0 +1,159 @@
...
+        System.out.println("Fine day we're having! I am Bartholomew, at your service!");
+        System.out.println("What may I do for you?");
+
+        Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> tasks = Storage.loadTasks();  // Load tasks from file
+
+        while (true) {
+            String input = scanner.nextLine().trim();
+
+            if (input.equalsIgnoreCase(EXIT_COMMAND)) {
```
NCF3535: There is code duplication here in the different cases:
`input.startsWith()` and others
You could try to use the DRY (Don't Repeat Yourself) principle to avoid this from happening, maybe use a switch will help as well.

src/main/java/Bartholomew.java~
NCF3535, line 1: Remove these unused files and use .gitignore so these files don't show up on GitHub. Alternatively, you may always use the version control to retrieve old versions of the code.

src/main/java/Storage.java
```diff
@@ -0,0 +1,76 @@
+import java.io.*;
```
NCF3535, line 1: Imported classes should be listed explicitly, even if you are using the entire class.
For example:
`import java.util.List;`
This gives an excellent documentation value for the class at hand and makes the class easier to comprehend and maintain.

NCF3535 (commented):
Generally the code really clean and easy to read, good job! You could further improve the code by reducing the amount of repetitions in the code and putting some of them into packages. This will help make it easier to read and understand. Keep up the good work!

PR #119, opened 2025-02-04
[xb] iP

xubin0:


src/main/java/xb.java
```diff
@@ -1,8 +1,12 @@
 import java.util.Scanner;
 
+
 public class xb {
+    public static final int MAX_LIST_SIZE = 100;
```
ecxm02: Good use of constants to avoid Magic Numbers. However, consider using a constants class to store all your constants so that it is easier to manage in the future when there are more constants

src/main/java/xb.java
```diff
@@ -1,8 +1,12 @@
 import java.util.Scanner;
 
+
 public class xb {
```
ecxm02: Class names should follow PascalCase, ie. "Xb". First letter should be capitalised

src/main/java/TaskList.java
```diff
@@ -1,15 +1,15 @@
 public class TaskList {
-    public static final int MAX_TASK_SIZE = 100;
     private static int taskCount;
```
ecxm02: This shouldn't be static, each instance of TaskList should have its own taskCount counter.

src/main/java/TaskPrinter.java
```diff
@@ -0,0 +1,33 @@
+public class TaskPrinter {
+    public static void printTasks(Task task) {
+        System.out.print(task.getTaskId() + ". ");
+        printTaskType(task);
+        printIsDone(task);
+        System.out.println(task.getTaskName());
+    }
+
+    public static void printIsDone(Task task) {
+        if (task.getIsDone()) {
```
ecxm02: (Can't find where the actual method is, so I'm referencing a call to that method). But, although it is a getter, perhaps it makes more sense to call the method "isDone()" instead. Therefore it should look like "if (task.IsDone())", allowing for better understanding.

ecxm02 (commented):
Overall, good job! Just added a few comments on style and convention

src/main/java/Xb.java
```diff
@@ -0,0 +1,82 @@
+import java.util.Scanner;
+
+
+public class Xb {
+    public static final int MAX_LIST_SIZE = 100;
+
+
+    public static void main(String[] args) {
+        TaskList taskList = new TaskList(MAX_LIST_SIZE);
+        Scanner s = new Scanner(System.in);
```
IgoyAI: Scanner s = new Scanner(System.in);

Consider to rename it to something like scanner to clearly indicate its purpose and more semantic

src/main/java/Xb.java
```diff
@@ -0,0 +1,82 @@
...
+
+        System.out.println("Hello, I am xb!");
```
IgoyAI: Consider to use higher level of abstraction such as using PrintWelcomeMessage()

src/main/java/TaskList.java
```diff
@@ -0,0 +1,110 @@
...
+            } else {
+                tasks[taskId - 1].setIsDone(false); //if valid and marked, unmark it
+                System.out.println(tasks[taskId - 1].getTaskName() + " is unmarked");
+            }
+        } else {
+            System.out.println("no such task"); //not valid print error
+        }
+    }
+
+    public void renameTask(int taskID, String newTaskName) {
```
IgoyAI: Consider to rename it to taskId for consistency with the rest of your code

src/main/java/TaskList.java
```diff
@@ -0,0 +1,110 @@
...
+        if (taskID < taskCount) { //check if taskId valid
```
IgoyAI: Consider to adjust the condition to ensure consistency. If tasks are numbered starting at 1, then a valid task should satisfy

IgoyAI (commented):
Overall good!

src/main/java/Xb.java
```diff
@@ -0,0 +1,67 @@
...
+        TaskList taskList = loadTasks();
+        Scanner s = new Scanner(System.in);
+
+        System.out.println("Hello, I am xb!");
+        System.out.println("What is your name?");
+        String name = s.nextLine();
+        System.out.print(name);
+        System.out.println(", what can I do for you?");
+        String command;
+        CommandHandler c = new CommandHandler(taskList);
```
irving11119: I think as mentioned as above, try to avoid single character names for your variables for clarity.

src/main/java/Xb.java
```diff
@@ -0,0 +1,67 @@
...
+
+        boolean run = true;
+        while (run) {
+            command = s.nextLine().trim();
+            CommandParser cp = new CommandParser(command);
+            CommandChecker cc= new CommandChecker(cp.getParts());
```
irving11119: Minor nitpick but be consistent with your spacing around assignment operator

src/main/java/command/CommandList.java
```diff
@@ -0,0 +1,121 @@
+package command;
+
+import exception.XbException;
+import task.*;
```
irving11119: Try not to use wildcard imports for your import statements. Instead you should import everything you need individually

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public int getTaskCount() {
+        return tasks.toArray().length;
+    }
+
+
+    public void markTaskAsDone(int taskId) {
+        if (taskId <= tasks.toArray().length) { //check if taskId is valid
+            if (tasks.get(taskId - 1).getIsDone()) { //check if task is already unmarked
+                System.out.println(tasks.get(taskId - 1).getTaskName() + " was already marked as done");
```
irving11119, line 108: Perhaps when printing messages such as this, you could first format your string and then print it for readability

src/main/java/command/CommandList.java
```diff
@@ -0,0 +1,121 @@
...
+                System.out.println(e.getMessage());
+            }
+
+        }
+    },
+    EVENT {
+        @Override
+        public void execute(String[] parts, TaskList taskList) {
+            try {
+                if (parts.length == 2 && parts[1].contains("/from") && parts[1].contains("/to")) {
```
irving11119: For these types of conditional statements, I could suggest simplifying them for readability

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    public void deleteTask(int taskId) {
+        if (taskId <= tasks.toArray().length && taskId > 0) {
+            Task task = tasks.get(taskId - 1);
+            tasks.remove(taskId - 1);
+            System.out.println("deleted task: ");
+            task.printTask();
+            System.out.println();
```
irving11119, line 145: I would suggest grouping similar statements together and separating different groups using whitespaces. This can help make it more readable for code reviewers

src/main/java/Xb.java
```diff
@@ -0,0 +1,67 @@
+import command.*;
```
irving11119: Try not to use wildcard imports for your import statements. Instead you should import everything you need individually

src/main/java/command/CommandHandler.java
```diff
@@ -0,0 +1,20 @@
...
+import task.TaskList;
+
+public class CommandHandler {
+    private TaskList taskList;
+
+    public CommandHandler(TaskList taskList) {
+        this.taskList = taskList;
+    }
+
+    public boolean runCommand(String[] parts) {
```
irving11119: Take note that boolean methods should have boolean naming (e.g. hasCommand rather than runCommand)

irving11119 (commented):
I think you can look at using more descriptive variables for some of your classes. Additionally, take note of the coding standards violations and code quality suggestions I have made to improve readability

PR #122, opened 2025-02-04
[Quan Teng Wai] iP

adoorknob:


src/main/java/Duck.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+        taskList.get(taskId - 1).unmarkAsCompleted();
+        System.out.println("Ok, I've marked this task as not done yet:\n  [ ] " + taskList.get(taskId - 1).getTitle());
+        System.out.println(divider);
+    }
+
+    private static void printList() {
+        System.out.println("Here are the tasks in your list: ");
+        for (int i = 0; i < taskList.size(); i++) {
+            System.out.println((i + 1) + ". [" + (taskList.get(i).isCompleted() ? "X" : " ") + "] " + taskList.get(i).getTitle());
```
terryasdf: Maybe you could break this line into two for a better view.

src/main/java/Duck.java
```diff
@@ -0,0 +1,164 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Duck {
```
kagiura: great name 🦆

src/main/java/Duck.java
```diff
@@ -0,0 +1,164 @@
...
+            + "|  __  \\_   _  ___| | __\n"
+            + "| |  | | | | |/ __| |/ /   _\n"
+            + "| |__| | |_| | (__|   <  <(.)___\n"
+            + "|_____/ \\__,_|\\___|_|\\_\\  (____/\n";
+    static private ArrayList<Task> taskList = new ArrayList<Task>();
+    static private int taskCounter = 0;
+
+    public static void main(String[] args) {
+        printWelcomeMessage();
+        while (true) {
```
kagiura: i love the extraction of this code into its own function, great way to keep it simple

src/main/java/Duck.java
kagiura: i lowkey am wondering if you can just change the logo to a literal ascii art of a duck

src/main/java/Duck.java
```diff
@@ -0,0 +1,164 @@
...
+        System.out.println("What can I do for you?");
+        System.out.println(divider);
+    }
+
+    private static void processUserInput() {
+        Scanner scanner = new Scanner(System.in);
+        String input = scanner.nextLine();
+        System.out.println(divider);
+
+        if (input.equalsIgnoreCase("bye")) {
```
kagiura: perhaps this could be changed to a switch statement? eg.
```
command = input.toLowercase();
switch (command) {
  case "xxx": // ...
  case "yyy": // ...
}```
so its more scalable in the future

File: src/main/java/Duck.java [hunk 5]
```diff
@@ -0,0 +1,164 @@
...
+
+    private static void printList() {
+        System.out.println("Here are the tasks in your list: ");
+        for (int i = 0; i < taskList.size(); i++) {
+            System.out.println((i + 1) + ". [" + (taskList.get(i).isCompleted() ? "X" : " ") + "] " + taskList.get(i).getTitle());
+        }
+        System.out.println(divider);
+    }
+
+    private static void printTalkingDuckWithLeftMargin(int margin) {
```
kagiura: great function name

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+    public boolean isCompleted() {
+        return completed;
+    }
+
+    public void markAsCompleted() {
+        this.completed = true;
+    }
+
+    public void unmarkAsCompleted() {
```
kagiura: perhaps "markAsIncomplete()"? so 1) both methods shows up when you type "markAs..." in your IDE 2) the wording negates at a more logical spot

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
+    private String title;
+    private boolean completed = false;
+
+    public Task() {
+        this("null");
```
kagiura: i think a more intuitive default name could be better like "Untitled Task", unless your intention is to use this to indicate nullish...?

kagiura (commented):
love how this code is structured! 🦆

src/main/java/Task.java
adoorknob: true! i left the IDE to make it's own constructor and didn't notice this, thanks!

src/main/java/Task.java
adoorknob: 🫡 makes sense

src/main/java/Duck.java
adoorknob: that's a great point thank you

src/main/java/Duck.java
adoorknob: 🦆

terryasdf (commented):
Well done.

src/main/java/Duck.java
adoorknob: sounds good!

src/main/java/duke/Duck.java
```diff
@@ -0,0 +1,332 @@
...
+    private static Task addTodo(String input) throws EmptyTodoException {
+        if (input.isEmpty()) {
+            throw new EmptyTodoException();
+        }
+        return new Todo(input.trim());
+    }
+
+    private static Task addDeadline(String input) {
+        int byDateIndex = input.indexOf("/by");
+        if (byDateIndex == -1) {
```
aditichadha1310: Avoid using magic strings and magic numbers.

aditichadha1310 (commented):
excellent work!

PR #123, opened 2025-02-04
[TiangSoonYong] iP

TiangSoonYong:


src/main/java/Doraemon.java
```diff
@@ -0,0 +1,74 @@
+import java.util.Scanner;
+
+
+public class Doraemon {
+    // Will go to Formatter class in future development
```
ShengBin-101: I like how you give comments to remind yourself and inform others of future development, this helps others understand your code design. I noticed this is other parts of your codes, good job!

src/main/java/Doraemon.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String MESSAGE_GOODBYE = "Bye. Hope to see you again soon!";
+    private static final String MESSAGE_HELLO = "Hello! I'm Doraemon!\n\t What can I do for you?";
+    private static final String COMMAND_BYE = "bye";
+    private static final String COMMAND_LIST = "list";
+    private static final String COMMAND_MARK = "mark";
+    private static final String COMMAND_UNMARK = "unmark";
+
+    private static final String COMMAND_TODO = "todo";
+    private static final String COMMAND_DEADLINE = "deadline";
+    private static final String COMMAND_EVENT = "event";
```
ShengBin-101: I like how you made use of constants to define the commands and common messages that are not meant to be mutable. The naming of the constants adhere to coding standard and is clear in telling others the use of the constant.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,77 @@
...
+        final int indexOfToPrefix = commandArgs.indexOf(DATA_PREFIX_TO);
+        int indexOfFirstPrefix = Math.min(indexOfFromPrefix, indexOfToPrefix);
+        String description;
+        String from;
+        String to;
+
+        description = commandArgs.substring(0, indexOfFirstPrefix).trim();
+        if (indexOfFromPrefix < indexOfToPrefix) { // Description - From - To
+            from = removePrefixSign(commandArgs.substring(indexOfFromPrefix, indexOfToPrefix), DATA_PREFIX_FROM).trim();
+            to = removePrefixSign(commandArgs.substring(indexOfToPrefix, commandArgs.length()), DATA_PREFIX_TO).trim();
```
ShengBin-101: I like that your function clearly describes its own purpose.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,77 @@
+public class TaskManager {
```
ShengBin-101: Pascal Casing for class names, adheres to coding standard.

src/main/java/TaskType.java
```diff
@@ -0,0 +1,3 @@
+public enum TaskType{
+    TODO, DEADLINE, EVENT
```
ShengBin-101: Enums that represent specific task types, useful for further addition of other task types in the future. Adheres to coding standard.

ShengBin-101 (commented):
Overall, great work with adhering to the coding standards specified. 

- I like how your code includes comments to tell others your future development which helps to convey your code design. 
- Class/Variable/Enum naming follows the coding standard. 
- Functions are well named to convey the use of the function in one glance.

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public void setDone(boolean done) {
```
Phua-Lock-Hian: the function parameter being done instead of isDone might be misleading, maybe just align with what is existing and use isDone.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,77 @@
...
+        String to;
+
+        description = commandArgs.substring(0, indexOfFirstPrefix).trim();
+        if (indexOfFromPrefix < indexOfToPrefix) { // Description - From - To
+            from = removePrefixSign(commandArgs.substring(indexOfFromPrefix, indexOfToPrefix), DATA_PREFIX_FROM).trim();
+            to = removePrefixSign(commandArgs.substring(indexOfToPrefix, commandArgs.length()), DATA_PREFIX_TO).trim();
+
+        } else { // Description - To - From
+            from = removePrefixSign(commandArgs.substring(indexOfFromPrefix, commandArgs.length()), DATA_PREFIX_FROM).trim();
+            to = removePrefixSign(commandArgs.substring(indexOfToPrefix, indexOfFromPrefix), DATA_PREFIX_TO).trim();
```
Phua-Lock-Hian: might want to split some lines here, it's getting a bit long and difficult to read.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,77 @@
...
+        String message = "Here are the tasks in your list:";
+        for (int i = 0; i < taskCount; i++) {
+            message += "\n\t " + String.format("%d. ", i + 1) + tasks[i];
+        }
+        return message;
+    }
+
+    public String markDone(int taskIndex, boolean done) {
+        tasks[taskIndex].setDone(done);
+        return "Nice! I've marked this task as " + (done ? "done" : "not done yet") + ":\n\t\t " + tasks[taskIndex];
```
Phua-Lock-Hian: the code may be more readable if you move the if-else expression outside before printing.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public String getTasks() {
+        String message = "Here are the tasks in your list:";
+        for (int i = 0; i < taskCount; i++) {
+            message += "\n\t " + String.format("%d. ", i + 1) + tasks[i];
+        }
+        return message;
+    }
+
+    public String markDone(int taskIndex, boolean done) {
```
Phua-Lock-Hian: this seems like we can only mark tasks as done but the required functionality is to mark or unmark. the functionality looks fine so perhaps just a name change would be good to minimize possible confusion, e.g. setIsDone() instead of markDone().

src/main/java/Doraemon.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static final TaskManager TASK_MANAGER = new TaskManager();
+    private static final Scanner SCANNER = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        echo(MESSAGE_HELLO);
```
Phua-Lock-Hian: is echo() still needed or can you just print?

Phua-Lock-Hian (commented):
Looks good overall. I like the effort in documenting with comments and noting of future development plans. Personally I feel that there are too many constants used that I have to keep referring back to.

src/main/java/Doraemon.java
TiangSoonYong: Yes, I use `echo(String... message)` to simplify the process of printing the top and bottom `LINE_DIVIDER` along with displayed message. Additionally, I plan to push all printings to be done by a `Formatter` object, so that my main() will be free of `System.out.println()`.

src/main/java/Task.java
TiangSoonYong: Thank you for spotting this error, I shall get it rectified 😄

src/main/java/TaskManager.java
TiangSoonYong: Agreed! I will create newlines for better code clarity.

src/main/java/TaskManager.java
TiangSoonYong: Agreed, it made the code too long as well!

TiangSoonYong (commented):
Thanks you for the comments! I will rectify the issues and push the updated commit. All the best for your iP too!

src/main/java/Doraemon.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String COMMAND_LIST = "list";
+    private static final String COMMAND_MARK = "mark";
+    private static final String COMMAND_UNMARK = "unmark";
+    private static final String COMMAND_TODO = "todo";
+    private static final String COMMAND_DEADLINE = "deadline";
+    private static final String COMMAND_EVENT = "event";
+    private static final String MESSAGE_INVALID = "Invalid Command";
+
+    private static final TaskManager TASK_MANAGER = new TaskManager();
+    private static final Scanner SCANNER = new Scanner(System.in);
```
zavsky: Should `TASK_MANAGER` and `SCANNER` be camelCase as they are not constants but rather object names?

src/main/java/Doraemon.java
```diff
@@ -0,0 +1,74 @@
...
+            return TASK_MANAGER.setIsDone(Integer.parseInt(commandArgs) - 1, false);
+        case COMMAND_TODO:
+            return TASK_MANAGER.addTask(commandArgs, TaskType.TODO);
+        case COMMAND_DEADLINE:
+            return TASK_MANAGER.addTask(commandArgs, TaskType.DEADLINE);
+        case COMMAND_EVENT:
+            return TASK_MANAGER.addTask(commandArgs, TaskType.EVENT);
+        default:
+            // Invalid command should throw an exception but assume all valid for now
+            return MESSAGE_INVALID;
```
zavsky: Should the cases each have a `break` statement or a `// Fallthrough` comment here?

src/main/java/Doraemon.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    private static String[] splitCommandWordAndArgs(String rawUserInput) {
+        final String[] split = rawUserInput.trim().split("\\s+", 2);
+        return split.length == 2 ? split : new String[]{split[0], ""}; // else case: no parameters
```
zavsky: Maybe consider splitting this line up as it could be confusing to read?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,86 @@
...
+            // Throw exception
+        }
+        taskCount++;
+        return "Got it. I've added this task:" +
+                "\n\t\t" + tasks[taskCount - 1] +
+                "\n\t Now you have " + taskCount + " tasks in the list.";
+    }
+
+    private void addToDo(String commandArgs) {
+        String description = commandArgs;
```
zavsky: Is this necessary? Or is this in preparation for checking if the input is valid?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,86 @@
...
+        tasks[taskCount] = new ToDo(description);
+    }
+
+    private void addDeadline(String commandArgs) {
+        final int indexOfByPrefix = commandArgs.indexOf(DATA_PREFIX_BY);
+        String description = commandArgs.substring(0, indexOfByPrefix).trim();
+        String by = removePrefixSign(commandArgs.substring(indexOfByPrefix, commandArgs.length()), DATA_PREFIX_BY).trim();
```
zavsky: Is this necessary or could the prefix be skipped completely by using `indexOfByPrefix + 1`?

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,86 @@
...
+        String from;
+        String to;
+
+        description = commandArgs.substring(0, indexOfFirstPrefix).trim();
+        if (indexOfFromPrefix < indexOfToPrefix) { // Description - From - To
+            from = removePrefixSign(
+                    commandArgs.substring(indexOfFromPrefix, indexOfToPrefix),
+                    DATA_PREFIX_FROM).trim();
+            to = removePrefixSign(commandArgs.substring(indexOfToPrefix, commandArgs.length()),
+                    DATA_PREFIX_TO).trim();
```
zavsky: I like your implementation to check the order of the user commands given to make sure that the program takes in the correct parameters

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,86 @@
...
+        String message = "Here are the tasks in your list:";
+        for (int i = 0; i < taskCount; i++) {
+            message += "\n\t " + String.format("%d. ", i + 1) + tasks[i];
+        }
+        return message;
+    }
+
+    public String setIsDone(int taskIndex, boolean isDone) {
+        tasks[taskIndex].setDone(isDone);
+        if (isDone) {
```
zavsky: I like that your code follows the conventions very well

zavsky (commented):
Your code is very readable and overall a very good job! Not sure if the `taskCount` updates properly though as it does not seem to increment after task creation. Just a slight nit

src/main/java/Doraemon.java
TiangSoonYong: Both objects were actually declared as a constant with `static final`, thus the naming convention

src/main/java/Doraemon.java
TiangSoonYong: I am unsure whether a `break` statement is required as each cases ends with a `return`

src/main/java/Doraemon.java
TiangSoonYong: Agreed! I will change it to individual if-else expression

src/main/java/TaskManager.java
TiangSoonYong: Actually, I left it as such to follow the same abstraction level. But it seems like I can use it for throwing exception too. Thank you for pointing that out!

src/main/java/TaskManager.java
TiangSoonYong: Personally, using the `removePrefixSign` method seems easier for me as it is able to handle any prefixes and I do not need to consider the number of chars to skip. Perhaps another method would be `indexOfByPrefix + DATA_PREFIX_BY.length()`

TiangSoonYong (commented):
Thank you for your comments once again! I will make any relevant changes. All the best!

src/main/java/doraemon/Doraemon.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Scanner;
+
+
+public class Doraemon {
+    // Will go to Formatter class in future development
+    private static final String LINE_PREFIX = "\t ";
+    private static final String LINE_DIVIDER = "__________________________________________________";
+
+    // Will go to Parser class in future development
+    private static final String LS = "\n\t "; // Line Separator
```
TVageesan: Name this LINE_SEPERATOR. In general, avoid abbreviations. This [video](https://youtu.be/-J3wNP6u5YU?t=69) has a really good explanation at to why (just the 1:00-2:00 part)

src/main/java/doraemon/task/TaskManager.java
```diff
@@ -0,0 +1,150 @@
...
+            return "The deadline is empty!";
+        } catch (NoFromPrefixException e) {
+            return "From prefix [/from] is missing!";
+        } catch (NoFromStringException e) {
+            return "The start date is missing!";
+        } catch (NoToPrefixException e) {
+            return "To prefix [/to] is missing!";
+        } catch (NoToStringException e) {
+            return "The end date is missing!";
+        }
```
TVageesan: Good idea to use Exceptions to simplify error handling! You could make this code even simpler by making the exceptions all inherit from a common exception (maybe MissingArgumentException), then just have a generic `catch (MissingArgumentException e) { return e.message }`.

src/main/java/doraemon/task/TaskManager.java
```diff
@@ -0,0 +1,150 @@
...
+        String by = removePrefixSign(commandArgs.substring(indexOfByPrefix, commandArgs.length()), DATA_PREFIX_BY).trim();
+        if (by.isEmpty()) {
+            throw new NoByStringException();
+        }
+        tasks[taskCount] = new Deadline(description, by);
+    }
+
+    private void addEvent(String commandArgs)
+            throws NoDescriptionException, NoFromPrefixException, NoFromStringException,
+            NoToPrefixException, NoToStringException {
```
TVageesan: Following from above, having the generic exception would also simplify declarations like this to `throws MissingArgumentException`

src/main/java/doraemon/task/TaskManager.java
```diff
@@ -0,0 +1,150 @@
...
+        String message = "Here are the tasks in your list:";
+        for (int i = 0; i < taskCount; i++) {
+            message += "\n\t " + String.format("%d. ", i + 1) + tasks[i];
+        }
+        return message;
+    }
+
+    public String setIsDone(int taskIndex, boolean isDone) {
+        try {
+            tasks[taskIndex].setDone(isDone);
```
TVageesan: Instead of directly manipulating the tasks array, `tasks[taskIndex]` consider putting it into a suitable variable first. Here, it could be `selectedTask = tasks[taskIndex]; selectedTask.setDone(isDone)`. this makes it easier to read and understand the code as the reader doesnt have to understand what tasks array is or what data it holds.

TVageesan (commented):
Overall, really well done. Honestly couldn't find much wrong readability wise with your code, and your future implementations of Exceptions/Formatter/Parser are solid design decisions. Left some nitpicks and suggestions to further improve.

src/main/java/doraemon/Doraemon.java
TiangSoonYong: That's a really good point that I did not consider!

src/main/java/doraemon/task/TaskManager.java
TiangSoonYong: Oh wow, that is a great idea. That way, I won't have to keep adding on more catches for further developments!

src/main/java/doraemon/task/TaskManager.java
TiangSoonYong: I see, thank you for this suggestion!

src/main/java/doraemon/task/TaskManager.java
TiangSoonYong: I like your suggestion!

TiangSoonYong (commented):
Thank you for the comments and suggestions. They were great and I will implement them in the next commit!

PR #124, opened 2025-02-04
[Shanice Tan Hui] iP

shanicetanhui:


src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+
+        while (true) {
+            String input = sc.nextLine();
+            String[] inputParts = input.split(" ", 2);
+
+            // Exit if user types "bye"
+            if (input.equals("bye")) {
+                System.out.println("____________________________________________________________");
```
szeyingg: might be better to use a named constant for the lines instead

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
+import java.util.Scanner;
+
 public class Duke {
     public static void main(String[] args) {
```
szeyingg: might be helpful to split parts of main code into different methods, easier for readers and for you to debug

szeyingg: maybe can rename file and class name to match name of chatbot

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+            // Marking tasks
+            } else if (inputParts[0].equals("mark")) {
+                int taskNumber = Integer.parseInt(inputParts[1]) - 1;
+                tasks[taskNumber].markAsDone();
+                System.out.println("____________________________________________________________");
+                System.out.println(" Nice! I've marked this task as done:");
+                System.out.println("   " + tasks[taskNumber]);
+                System.out.println("____________________________________________________________");
+
+            } else if (inputParts[0].equals("unmark")) {
```
szeyingg: might be clearer and easier if you use switch case instead of many if-else statements?

szeyingg (commented):
looks good, just minor changes to improve readability :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+                System.out.println("____________________________________________________________");
+
+            // Event tasks
+            } else if (inputParts[0].equals("event")) {
+                String[] parts = inputParts[1].split("/from");
+                String description = parts[0].trim();
+                String[] dateParts = parts[1].split("/to");
+                String from = dateParts[0].trim();
+                String to = dateParts[1].trim();
+                tasks[taskCount] = new Event(description, from, to);
```
Lukapeng77: For the To do, Deadline and Event tasks, actually some of the print methds can be reusable, can consider use a public method to do the add task function and make use of the Polymorphism technique.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+                System.out.println("____________________________________________________________");
+
+            // To Do tasks
+            } else if (inputParts[0].equals("todo")) {
+                String description = inputParts[1];
+                tasks[taskCount] = new ToDo(description);
+                taskCount++;
+
+                System.out.println("Got it. I've added this task:");
+                System.out.println(tasks[taskCount - 1]);
```
Lukapeng77: Can use another variable to store the taskCount -1 to illustrate the meaning of this variable, avoid use the magic number.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+        System.out.println(" Hello! I'm Pesto!");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+
+        while (true) {
+            String input = sc.nextLine();
```
Lukapeng77: It's better to use more meaningful variable names for the sc, like the UserInput.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+        System.out.println("____________________________________________________________");
+        System.out.println(" Hello! I'm Pesto!");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        Scanner sc = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+
+        while (true) {
```
Lukapeng77: For check the continue condition, can create a new variable to store and explain the meaning of the variable. 
Eg: Boolean isExist = false; To check for the continue condition.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+            // To Do tasks
+            } else if (inputParts[0].equals("todo")) {
+                String description = inputParts[1];
+                tasks[taskCount] = new ToDo(description);
+                taskCount++;
+
+                System.out.println("Got it. I've added this task:");
+                System.out.println(tasks[taskCount - 1]);
+                System.out.println("Now you have " + taskCount + " tasks in the list.");
+                System.out.println("____________________________________________________________");
```
RyanLing169: Quite a long line. Should it be wrapped?

src/main/java/Duke.java
srfl0: having another variable to store taskCount -1 would mean having to update it, I would think commenting for what it represents here, which is to map user-read values to program-read values caused by 0-indexing, works better here.

src/main/java/Duke.java
```diff
@@ -36,35 +36,77 @@ public static void main(String[] args) {
...
+
+            // Marking tasks
             } else if (inputParts[0].equals("mark")) {
                 int taskNumber = Integer.parseInt(inputParts[1]) - 1;
-                taskList[taskNumber].markAsDone();
+                tasks[taskNumber].markAsDone();
                 System.out.println("____________________________________________________________");
```
srfl0, line 86: possible to extract print statements like these into another method to lighten up main?

src/main/java/Duke.java
srfl0: having another variable to store taskCount -1 would mean having to update it, I would think commenting for what it represents here, which is to map user-read values to program-read values caused by 0-indexing, works better here.

src/main/java/Duke.java
```diff
@@ -36,35 +36,77 @@ public static void main(String[] args) {
...
+                System.out.println("Got it. I've added this task:");
+                System.out.println(tasks[taskCount - 1]);
+                System.out.println("Now you have " + taskCount + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+
+            // Marking tasks
             } else if (inputParts[0].equals("mark")) {
                 int taskNumber = Integer.parseInt(inputParts[1]) - 1;
-                taskList[taskNumber].markAsDone();
+                tasks[taskNumber].markAsDone();
```
srfl0: possible to extract print statements like these into another method to lighten up main?

src/main/java/Duke.java
```diff
@@ -1,10 +1,265 @@
+import java.util.Scanner;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class Duke {
+    private static final String FILE_PATH = "/Users/shanice/CS2113/ip/data/duke.txt"; // Relative path to save file
+
     public static void main(String[] args) {
```
Hackin7: This method is quite long, splitting it up would greatly improve readability

src/main/java/Duke.java
```diff
@@ -1,10 +1,265 @@
...
+            }
+
+            private static int loadTasks (Task[]tasks){
+                int taskCount = 0;
+                try {
+                    File file = new File(FILE_PATH);
+                    if (!file.exists()) {
+                        System.out.println("No previous task data found. Starting with a new task list.");
+                        return taskCount;  // No file, so return with taskCount = 0
+                    } /* else {
```
Hackin7: Remove commented code

Hackin7 (commented):
More abstraction and SLAP

src/main/java/Duke.java
```diff
@@ -1,10 +1,265 @@
+import java.util.Scanner;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class Duke {
```
Hackin7: Should ideally rename Duke to the name of your chatbot

PR #125, opened 2025-02-04
[Wong Chuan Hui] iP

chuanhuiw:


src/main/java/cuckoo.java
```diff
@@ -0,0 +1,8 @@
+public class cuckoo {
```
xiaojielei: Good to change the name of class to make it specialized

src/main/java/cuckoo.java
```diff
@@ -1,14 +1,78 @@
 import java.util.Scanner;
 
+class Task {
```
xiaojielei: Great! You split the code into seperate methods

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+        String chatbotName = "cuckoo";
+        System.out.println("ANNYEONG! I am " + chatbotName);
+        System.out.println("Input 'list' to view existing list or add a new Todo");
+        System.out.println("____________________________________________________________");
+        Scanner in = new Scanner(System.in);
+
+        while (true) {
+            String input = in.nextLine();
+
+            if (input.equalsIgnoreCase("bye")) {
```
xiaojielei: The if-else statement is clear

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
+import java.util.Scanner;
+
+class Task {
+    protected String description;
```
xiaojielei: clear indentation

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+        return (isDone ? "X" : " ");
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
```
Deanson-Choo: Is 'this.' necessary? Perhaps you can just write isDone = true for example since there are no conflicts.

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public String toString() {
+        return "[" + getStatus() + "] " + description;
+    }
+}
```
Deanson-Choo: Good usage of the Override label!

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+
+class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+    }
+}
+
+public class cuckoo {
```
Deanson-Choo: Please write all your classes' names in PascalCase, so 'Cuckoo' instead of 'cuckoo'.

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+    private static Task[] tasks = new Task[100];
+    private static int taskCount = 0;
+
+    public static void listTasks() {
+        System.out.println("___________________________________________________________");
```
Deanson-Choo: Consider assigning the 'horizontal line' to a variable so you can refer to it easily in your code.

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+            tasks[index - 1].markAsNotDone();
+            System.out.println("___________________________________________________________");
+            System.out.println("Ok, I've marked this task as not done yet:");
+            System.out.println(" " + tasks[index - 1]);
+            System.out.println("___________________________________________________________");
+        }
+    }
+
+    public static void main(String[] args) {
+        String chatbotName = "cuckoo";
```
Deanson-Choo: Maybe you can consider adding the name of the chatbot under the class Cuckoo

src/main/java/cuckoo.java
```diff
@@ -0,0 +1,106 @@
...
+            } else {
+                if (taskCount < 100) {
+                    tasks[taskCount++] = new Todo(input);
+                    System.out.println("____________________________________________________________");
+                    System.out.println(" added: " + input);
+                    System.out.println("____________________________________________________________");
+                } else {
+                    System.out.println(" Task list is full!");
+                }
+            }
```
Deanson-Choo: Maybe you can consider adding another class to do all these condition checking in order to neaten the code in main. Also, it would be good to add comments for each condition to better understand the flow of the program.

Deanson-Choo (commented):
Good job overall! Just a few touchups will make it even better!

src/main/java/Parser/Parser.java
```diff
@@ -0,0 +1,60 @@
...
+        } else if (input.startsWith("unmark ")) {
+            tasks.unmarkTask(parseIndex(input, 7));
+        } else if (input.startsWith("delete ")) {
+            tasks.deleteTask(parseIndex(input, 7));
+        } else if (input.startsWith("find ")) {
+            FindCommand.execute(tasks, input.substring(5).trim());
+        } else {
+            throw new macdonaldException("Unknown command! Use 'todo', 'deadline', 'event', 'mark', 'unmark', 'delete', or 'list'.");
+        }
+    }
```
yijiano, line 44: Could have this been done using switch/case instead?

yijiano (commented):
Good work

src/main/java/Parser/Parser.java
chuanhuiw: > Could have this been done using switch/case instead?

can is can but if im not wrong Java's switch doesn't support complex conditions like startsWith() so if I implement switch/case i'll need to extract command then use switch

PR #126, opened 2025-02-04
[Wong Man Zhong] iP

ManZ9802:


src/main/java/Jerry.class
Betahaxer, line 1: hey should .class files be pushed?

src/main/java/Jerry.java
```diff
@@ -0,0 +1,141 @@
...
+    public static void printTab(String message) {
+        System.out.println("\t" + message);
+    }
+
+    public static void printHoriLine() {
+        for (int i = 50; i > 0; i--) {
+            System.out.print("_");
+        }
+        System.out.println();
+    }
```
Betahaxer: Should horizontal be shortened in the name here?

src/main/java/Jerry.java
```diff
@@ -0,0 +1,141 @@
+import java.util.Scanner;
+
+public class Jerry {
+    public static void printTab(String message) {
+        System.out.println("\t" + message);
+    }
```
Betahaxer: Can the naming of this function be improved to show what the function does? Even if it is longer?

src/main/java/Jerry.java
```diff
@@ -0,0 +1,141 @@
...
+                newEntry(list, event);
+                printNewEntry(list, event);
+                break;
+
+            default:
+                printHoriLine();
+                printTab("please define task: " + text);
+                printHoriLine();
+                break;
+            }
```
Betahaxer: Good use of switch case here to split the input based on the commands!

Betahaxer (commented):
Overall, the classes were well structured and the naming of the classes, methods and attributes are quite appropriate. With some minor changes to the code, the project would be  good to go.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,21 @@
...
+        return by;
+    }
+
+    public void setBy(String by) {
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
```
yangyi-zhu, line 23: Excellent use of inheritance and overriding methods!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,21 @@
+public class Deadline extends Task {
+    protected String by;
```
yangyi-zhu, line 6: Would suggest a slightly clearer variable name. Its functionality does not seem clear to reviewers prior to reading its implementation below.

src/main/java/Task.java
```diff
@@ -19,7 +19,8 @@ public void markAsNotDone() {
         isDone = false;
     }
 
-    public void printTask() {
```
yangyi-zhu: Good implementation that removes redundant methods!

src/main/java/Jerry.java
```diff
@@ -1,4 +1,44 @@
...
 
 public class Jerry {
-}
+    public static void printHoriLine() {
+        System.out.println("\t");
+        for (int i = 50; i > 0; i--) {
+            System.out.print("_");
+        }
+        System.out.println("");
+    }
```
yangyi-zhu: Methods like this can be replaced by simply using constants.
`public static final String DIVIDER = "__________________________________________________";`

src/main/java/Jerry.java
```diff
@@ -86,18 +95,41 @@ public static void main(String[] args) {
...
+                String eventDescription = text.substring(6, text.indexOf('/')).trim();
+                int firstIndex = text.indexOf('/');
+                int secondIndex = text.indexOf('/', firstIndex + 1);
+                String start = text.substring(firstIndex + 6, secondIndex).trim();
+                String end = text.substring(secondIndex + 4).trim();
+                Event event = new Event(eventDescription, start, end);
+                newEntry(list, event);
+                printNewEntry(list, event);
+                break;
+
```
yangyi-zhu: The main method looks very large and not very organized. 76 lines is a bit too long for the main method.

yangyi-zhu (commented):
Overall, I think this works fine. It however is quite difficult to read sometimes and the presentation can see some improvements. 

On a side note, in terms of version control, I think it is best to name the commits that suggest what changes have been made as it is difficult to backtrack on what exactly you changed each time. The increments for CS2113 should be added as tags.

src/main/java/CommandManager.java
```diff
@@ -0,0 +1,268 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Objects;
+
+public class CommandManager {
+    public static void indentMessage(String message) {
+        System.out.println("\t" + message);
+    }
+
+    public static void exitProcess(ArrayList<Task> list) throws IOException {
```
sevenseasofbri: This function does not exactly do the work of exiting the process, rather, it prints an exit message and saves the data. Consider naming it something more apt.

src/main/java/CommandManager.java
```diff
@@ -0,0 +1,268 @@
...
+            String line = s.nextLine();
+            Task task = parseTask(line);
+            list.add(task);
+        }
+        s.close();
+    }
+
+    private static Task parseTask(String line) {
+        char taskType = line.charAt(1);
+        boolean isDone = line.charAt(4) == 'X';
```
sevenseasofbri: Avoid using magic literals by storing them in constants or variables

src/main/java/CommandManager.java
```diff
@@ -0,0 +1,268 @@
...
+        printHoriLine();
+        indentMessage("Added task to list:");
+        indentMessage(key.toString());
+        indentMessage("Now you have " + list.size() + " task(s)");
+        printHoriLine();
+    }
+
+    private static void markTask(ArrayList<Task> list, String text, boolean done) {
+        String number = text.replaceAll("\\D+", ""); // Remove all non-digits
+        int i = Integer.parseInt(number);
```
sevenseasofbri: The name `i` is not intuitive, maybe call it `taskNumber` instead. Additionally, the variable `number` can be called `parsedNumber` instead.

src/main/java/CommandManager.java
```diff
@@ -0,0 +1,268 @@
...
+        } else {
+            list.get(i - 1).markAsNotDone();
+            printHoriLine();
+            indentMessage("Marked task " + i + " as not done");
+        }
+        indentMessage(list.get(i - 1).toString());
+        printHoriLine();
+    }
+
+    private static void createAndAddTask(ArrayList<Task> list, String text, String type) throws IllegalDeadlineException, IllegalEventException {
```
sevenseasofbri: Lines longer than 120 characters should be wrapped to the next line

src/main/java/CommandManager.java
```diff
@@ -0,0 +1,268 @@
...
+        int i = Integer.parseInt(number);
+        printHoriLine();
+        indentMessage("Noted, deleted task " + i + ": ");
+        indentMessage(list.get(i - 1).toString());
+        indentMessage("Now you have " + (list.size() - 1) + " task(s)");
+        printHoriLine();
+        list.remove(i - 1);
+    }
+
+    public static boolean processCommand(ArrayList<Task> list, String text) {
```
sevenseasofbri, line 22: This function is very long (>30LOC). Consider breaking it into chunks of smaller functions

src/main/java/CommandManager.java
```diff
@@ -0,0 +1,268 @@
...
+                break;
+
+            case "unmark":
+                markTask(list, text, false);
+                break;
+
+            case "todo":
+            case "deadline":
+            case "event":
+                createAndAddTask(list, text, command);
```
sevenseasofbri: > From java coding standard: The explicit `// Fallthrough` comment should be included whenever there is a case statement without a break statement.

sevenseasofbri (commented):
Good job so far! Readability can be improved by reducing the length of some methods.

PR #127, opened 2025-02-04
[Jiang Yukuan] iP

jyukuan:


src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    //new class : task for store name and done or not
+    private static class Task {
```
J030104: It's always better to group classes of similar functionalities together and put them in different files.
Don't be afraid of switching between tabs. You can use go to definition/reference etc to help you navigate.
Get used to using hotkeys that help you find things. You'll pick up the skill very soon. Good luck.

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+        if (listCount < list.length) {
+            list[listCount] = new Task(input);
+            listCount++;
+            printIt("added: " + input);
+        } else {
+            printIt("Task list is full!");
+        }
+    }
+
+    private static void printList() {
```
J030104: You can try creating a class that deals with output formatting and avoid writing the "______" everywhere, so that your code can look cleaner

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+        System.out.println("____________________________________________________________");
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < listCount; i++) {
+            System.out.println((i + 1) + "." + list[i].getStatus() + " " + list[i].description);
+        }
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void markTaskAsDone(int index) {
```
J030104: Consider putting the methods like this one into the Task class, not in Duke. In this way, you or your teammate in the future can locate the bugs more easily and without getting lost in the project.

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+
+        void markNotDone() {
+            this.isDone = false;
+        }
+    }
+
+    private static Task[] list = new Task[100];
+    private static int listCount = 0;
+
+    private static void printIt(String message) {
```
tzqgav10: The method is not descriptive enough, would be better to be more specific with what the method is printing

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    //new class : task for store name and done or not
```
tzqgav10: Comments need to be grammatically correct

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+    private static Task[] list = new Task[100];
+    private static int listCount = 0;
+
+    private static void printIt(String message) {
+        System.out.println("____________________________________________________________");
+        System.out.println(message);
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void addToList(String input) {
```
tzqgav10: Method can be more specific like addTask()

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+            String input = scanner.nextLine();
+
+            if (input.equalsIgnoreCase("bye")) {
+                System.out.println("____________________________________________________________");
+                System.out.println("Bye. Hope to see you again soon!");
+                System.out.println("____________________________________________________________");
+                break;
+            } else if (input.equalsIgnoreCase("list")) {
+                printList();
+            } else if (input.length() > 5 && input.substring(0, 5).equals("mark ")) {
```
tzqgav10: Better to use input.startsWith() function to check if the input begins with the string that we want compared to substring.equals()

tzqgav10 (commented):
Code is functional but could be better written to follow coding guidelines

src/main/java/Duke.java
jyukuan: Thank you! I will note it.

src/main/java/Duke.java
jyukuan: got it

src/main/java/Duke.java
jyukuan: Thank you! I will note it.

src/main/java/Duke.java
jyukuan: Is input.startsWith("mark ", 1, 5) ?  Not sure about that function.

tzqgav10: It looks something like this:

if (line.startsWith("mark ")) {
  ...
}

You can explore and find out what works best!

jyukuan: Got it ! Thanks a lot!

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+                String numberPart = input.substring(5);
+                if (isValidNumber(numberPart)) {
+                    int index = Integer.parseInt(numberPart) - 1;
+                    markTaskAsDone(index);
+                } else {
+                    printIt("Invalid task number!");
+                }
+            } else if (input.length() > 7 && input.substring(0, 7).equals("unmark ")) {
```
J030104: Try to make the logic structure clearer.
Ref: https://nus-cs2113-ay2425s2.github.io/website/schedule/week4/topics.html#w4-6-code-quality-readability

J030104: Things like:
- Avoid deep nesting
- Avoid complicated expressions
- SLAP hard

are something you may want to know more about

src/main/java/Duke.java
jyukuan: Got it ! Thanks a lot!

src/main/java/Duke.java
jyukuan: Got it ! Thanks a lot!

src/main/java/Duke.java
jyukuan: Got it ! Thanks a lot!

src/main/java/Duke.java
jyukuan: Got it ! Thanks a lot!

src/main/java/Ui.java
```diff
@@ -0,0 +1,57 @@
...
+        System.out.println("Noted. I've removed this task:");
+        System.out.println("  " + task);
+        System.out.println("Now you have " + taskCount + " tasks in the list.");
+    }
+
+    public void showMarkStatus(Task task, boolean isDone) {
+        System.out.println("____________________________________________________________");
+        System.out.println("The task mark as" + (isDone ? "done" : "not done") + ":");
+        System.out.println("  " + task);
+        System.out.println("____________________________________________________________");
```
xenthm: I like the use of indentations and divider lines in your messages to make them more visually structured to users, but perhaps consider using a constant to represent it (i.e. `INDENT`) to avoid [magic numbers](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-magic-numbers).

src/main/java/TaskList.java
```diff
@@ -0,0 +1,112 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+public class TaskList {
+    private ArrayList<Task> tasks;
+    private static final Path DATA_PATH = Paths.get("data", "duke.txt");
+
+    public TaskList() {
+        this.tasks = new ArrayList<>();
```
xenthm: Consider omitting `this` in `this.tasks = ...` to reduce unnecessary noise since it is not shadowed by other variables in the constructor. This is according to [the Java coding standards for this course](https://se-education.org/guides/conventions/java/index.html#variables).

src/main/java/TaskList.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            // if file do not exist, just return
+            if (!Files.exists(DATA_PATH)) return;
+
+            List<String> lines = Files.readAllLines(DATA_PATH);
+            for (String line : lines) {
+                try {
+                    Task task = parseTask(line);
+                    if (task != null) tasks.add(task);
```
xenthm: While technically correct, consider adding braces to one-liner `if` statements. It helps prevent unintended effects if more statements are included in the `if` block. This is in accordance to the [course website](https://nus-cs2113-ay2425s2.github.io/website/schedule/week2/topics.html#c-to-java-control-flow-branching). The ideal format can be found in [the Java coding standards for this course](https://se-education.org/guides/conventions/java/index.html#layout) as well.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,112 @@
...
+                } catch (Exception e) {
```
xenthm: Good use of `try`-`catch` blocks to handle exceptions! However, do consider using a more specific subclass of the generic `Exception` to provide more context and potentially make debugging easier. For your case, you seem to already throw `IllegalArgumentException` in the `parseTask` method, so you can look into using that.

PR #130, opened 2025-02-04
[Tejas Kumar] iP

tejaskumar0:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,11 @@
+public class Deadline extends Task{
+
+    public Deadline(String description) {
+        super(description);
+    }
```
siqiangz, line 23: Add one line spacing following other task files, other than that looks great!

src/main/java/Tywin.java
```diff
@@ -0,0 +1,101 @@
...
+        Task[] tasks = new Task[100];
+        int taskCounter = 0;
+
+        Scanner in = new Scanner(System.in);
+
+        while (true) {
+            String line = in.nextLine();
+            String command = line.split(" ")[0]; // Extract the command
+
+            switch (command) {
```
siqiangz: should bring this switch statement out of the loop

src/main/java/Events.java
```diff
@@ -0,0 +1,12 @@
+public class Events extends ToDos{
```
siqiangz: looks good!!!

src/main/java/Tywin.java
```diff
@@ -0,0 +1,101 @@
...
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello I'm Tywin");
+        System.out.println("What can I do for you?");
+        System.out.println("----------------------");
+
+        Task[] tasks = new Task[100];
```
siqiangz: magic number: should bring this out as constant variable

siqiangz (commented):
Overall looks good, just a few small bits like spacing(maybe this is for readability my bad) and magic numbers. Also can put big switch statements in methods

src/main/java/Deadline.java
```diff
@@ -0,0 +1,11 @@
+public class Deadline extends Task{
```
juneja999: I like that you have used inheritance till now for various classes, it helps remove redundant parts.

src/main/java/Tywin.java
juneja999: I appreciate the case statement logic for finding key words, makes the code much more cleaner.

src/main/java/Tywin.java
```diff
@@ -0,0 +1,101 @@
...
+            case "event":
+                System.out.println("----------------------");
+                System.out.println("Added Event to list");
+                tasks[taskCounter] = new Events(line);
+                taskCounter++;
+                break;
+
+            case "mark":
+                try {
+                    int taskNumber = Integer.parseInt(line.split(" ")[1]) - 1;
```
juneja999: I think we can simplify this complex expression by dividing it into two lines.

src/main/java/Tywin.java
```diff
@@ -0,0 +1,101 @@
...
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello I'm Tywin");
+        System.out.println("What can I do for you?");
+        System.out.println("----------------------");
```
juneja999: Can we make "---------------------------" a constant and then print it?

juneja999 (commented):
LGTM. Mostly the code is simple to understand and the logic flows very well across the code.

src/main/java/Tywin.java
```diff
@@ -0,0 +1,132 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Tywin {
+    public static void main(String[] args) {
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
```
irving11119: You could potentially change this to a final static variable and use the naming conventions for magic strings

src/main/java/Tywin.java
```diff
@@ -0,0 +1,132 @@
...
+                    System.out.println("You have no more tasks! Enjoy your day!");
+                    break;
+                } else {
+                    System.out.println("Here are the tasks in your list:");
+                    for (int i = 0; i < tasks.size(); i++) {
+                        System.out.println((i + 1) + ". " + tasks.get(i));
+                    }
+                    System.out.println("----------------------");
+                    break;
+                }
```
irving11119: Your nesting here reaches 4 levels deep of nesting. Take note that for code readability, you should try to reduce the number of nestings to at most 3. You can abstract out your logic into methods to achieve this.

src/main/java/Tywin.java
```diff
@@ -0,0 +1,132 @@
...
+                        tasks.get(taskNumber).unmarkAsDone();
+                        System.out.println("OK, I've marked this task as not done yet:");
+                        System.out.println("  " + tasks.get(taskNumber));
+                    } else {
+                        System.out.println("Invalid task number.");
+                    }
+                } catch (Exception e) {
+                    System.out.println("Invalid input. Please specify a valid task number.");
+                }
+                System.out.println("----------------------");
```
irving11119: This print statement has been reused a couple of times. I think you can define this string as a constant (e.g. `divider = "----------------------"` and print the divider

src/main/java/Tywin.java
```diff
@@ -0,0 +1,132 @@
...
+                break;
+
+            default:
+                // Add a new task
+                System.out.println("⚠️ Error: Unrecognized command '" + command + "'.");
+                System.out.println("Please enter a valid command (todo, deadline, event, list, mark, unmark, bye,delete)");
+                System.out.println("----------------------");
+                break;
```
irving11119, line 18: This break is not necessary for the default case

src/main/java/Tywin.java
```diff
@@ -0,0 +1,132 @@
...
+                        System.out.println((i + 1) + ". " + tasks.get(i));
+                    }
+                    System.out.println("----------------------");
+                    break;
+                }
+
+            case "todo":
+                    if(parts.length < 2 || parts[1].trim().isEmpty()){
+                        System.out.println("Description cannot be empty");
+                        break;
```
irving11119: I think you can simplify this and have 1 break at the end of each case rather than having a break statement for each conditional block

src/main/java/Events.java
irving11119: I think in this case, a better naming convention for this would be `Event` rather than `Events`. Based on the coding standards, plural form should only be used to name objects when it represents a collection of objects (e.g. if your Obkect contains an array of events. However, in this case since it's representing one event, it's better to use the singular form. This applies to the `ToDos` class as well.

irving11119 (commented):
Take note of some of the coding standard violations as well as some feedback to improve code readability. Additionally, remember to gitignore the relevant folders (e.g. the out folder, META-INF, .idea) as these should not be committed and make it harder for code reviewers to review you code. If you have any issues with this feel free to reach out for help.

PR #131, opened 2025-02-04
[eth4n22] iP

eth4n22:


text-ui-test/input.txt
whalesyong, line 1: Great practice to start implementing tests using a file!

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
...
+
+        System.out.println("_____________________________");
+        System.out.println("Hello! I'm Baba");
+        System.out.println("What can I do for you?");
+        System.out.println("_____________________________");
+
+        while (true) {
+            String userInput = scanner.nextLine().trim();
+
+            if (userInput.equalsIgnoreCase("bye")) {
```
whalesyong: I think it'll be a good idea to use `switch` for this part of the code in future to check the user's first word / inputs. The `else if` operators will make the code much longer unnecessarily, especially if trying to add more functionality in the future.

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
...
+                if (parts.length < 2 || !parts[1].contains(" /to ")) {
+                    System.out.println("Invalid format. Use: event [task] /from [start] /to [end]");
+                } else {
+                    String[] timeParts = parts[1].split(" /to ", 2);
+                    tasks.add(new Event(parts[0], timeParts[0], timeParts[1]));
+                    System.out.println("_____________________________");
+                    System.out.println("Got it. I've added this task:");
+                    System.out.println(tasks.get(tasks.size() - 1));
+                    System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+                    System.out.println("_____________________________");
```
whalesyong: A suggestion would be to provide a simple function to print the line, instead of printing it directly to decrease clutter.

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
...
+                    System.out.println("Invalid format. Use: mark [task_number]");
+                }
+            } else if (userInput.startsWith("unmark ")) {
+                try {
+                    int taskIndex = Integer.parseInt(userInput.split(" ")[1]) - 1;
+                    if (taskIndex >= 0 && taskIndex < tasks.size()) {
+                        tasks.get(taskIndex).markAsNotDone();
+                        System.out.println("_____________________________");
+                        System.out.println("OK, I've marked this task as not done yet:");
+                        System.out.println(tasks.get(taskIndex));
```
whalesyong: In general, you could add more comments to your code, it'll make things a little easier to read.

whalesyong (commented):
All in all, I think things are good! I see some good practices, and the practices that need improvement are relatively minor :)

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Baba {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> tasks = new ArrayList<>();
```
xhuinn: Good that variable names such as `tasks`, `scanner` and` userInput` are clear and descriptive.

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
...
+                        System.out.println("Invalid task number.");
+                    }
+                } catch (Exception e) {
+                    System.out.println("Invalid format. Use: mark [task_number]");
+                }
+            } else if (userInput.startsWith("unmark ")) {
+                try {
+                    int taskIndex = Integer.parseInt(userInput.split(" ")[1]) - 1;
+                    if (taskIndex >= 0 && taskIndex < tasks.size()) {
+                        tasks.get(taskIndex).markAsNotDone();
```
xhuinn: Good that methods like `markAsDone()` and `markAsNotDone()` are well-named and self-explanatory.

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
...
+                System.out.println("_____________________________");
+                if (tasks.isEmpty()) {
+                    System.out.println("No tasks added yet.");
+                } else {
+                    System.out.println("Here are the tasks in your list:");
+                    for (int i = 0; i < tasks.size(); i++) {
+                        System.out.println((i + 1) + ". " + tasks.get(i));
+                    }
+                }
+                System.out.println("_____________________________");
```
xhuinn: Good that the divider stays cosistentent throughout the code. A suggestion would be to have a constant String DIVIDER instead of manually typing the divider repeatedly.

src/main/java/Baba.java
```diff
@@ -0,0 +1,104 @@
...
+                if (parts.length < 2) {
+                    System.out.println("Invalid format. Use: deadline [task] /by [date]");
+                } else {
+                    tasks.add(new Deadline(parts[0], parts[1]));
+                    System.out.println("_____________________________");
+                    System.out.println("Got it. I've added this task:");
+                    System.out.println(tasks.get(tasks.size() - 1));
+                    System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+                    System.out.println("_____________________________");
+                }
```
xhuinn: Instead of repeating these lines of code, perhaps it would be better to extract them into a helper method and call it

xhuinn (commented):
Generally code looks good and clear. 
Just a few nits to fix.

src/main/java/Baba.java
```diff
@@ -0,0 +1,155 @@
...
+        System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+        System.out.println("_____________________________");
+    }
+
+    private static void addDeadline(ArrayList<Task> tasks, String userInput) throws BabaException {
+        if (userInput.equals("deadline")) {
+            throw new BabaException("OOPS!!! The description of a deadline cannot be empty.");
+        }
+
+        String[] parts = userInput.substring(9).split(" /by ", 2);
```
TVageesan: The 9 here is a magic number. I would recommend either using another .split() instead of the hard index or putting the 9 into a constant.

src/main/java/Baba.java
```diff
@@ -0,0 +1,155 @@
...
+        if (parts[0].trim().isEmpty()) {
+            throw new BabaException("OOPS!!! The description of a deadline cannot be empty.");
```
TVageesan: Instead of directly using the parts array i.e. parts[0], it'd be more readable if it was put into a suitable variable such as `description`. This will make logic easier to read, i.e. `if description.isEmpty()` is much easier to parse on first glance. You can apply this general principle throughout your code to make it more readable.

src/main/java/Baba.java
```diff
@@ -0,0 +1,155 @@
...
+
+                if (userInput.equalsIgnoreCase("bye")) {
+                    System.out.println("_____________________________");
+                    System.out.println("Bye! Hope to see you again soon!");
+                    System.out.println("_____________________________");
+                    break;
+                } else if (userInput.equalsIgnoreCase("list")) {
+                    printTaskList(tasks);
+                } else if (userInput.startsWith("mark ")) {
+                    markTaskAsDone(tasks, userInput);
```
TVageesan: The strings "mark", "list" etc. are magic strings. Consider putting them into a constant or enum instead of direct string comparison. This applies to other string comparisons I noticed in your code as well.

TVageesan (commented):
Well done, just some minor nitpicks for better readability

PR #134, opened 2025-02-04
[James17042002] iP

James17042002:


src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+        int counter = 0;
+        greetings();
+
+        do {
+            line = in.nextLine();
+            switch (line.split(" ")[0]) {
+            case "list":
+                printTaskList(list, counter);
+                break;
+
```
nmarwah7: I like how you used a switch case here, it avoids deep nesting of if-then statements

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+public class Ava {
+
+    public static final String LINE_SEPARATOR = "_____________________________";
+    public static final int MAX_TASKS = 100;
+
+    public static void main(String[] args) {
+        String line;
+        Scanner in = new Scanner(System.in);
+        Task[] list = new Task[MAX_TASKS];
+        int counter = 0;
```
nmarwah7: Should this be given a more descriptive name perhaps, this could be a bit tricky to follow what its purpose is?

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+                handleMark(line, list, counter);
+                break;
+
+            case "unmark":
+                handleUnmark(line, list, counter);
+                break;
+
+            case "bye":
+                // Exit condition: just exit the loop
+                break;
```
nmarwah7: I like how you added a comment to improve understanding of this case, however should you think about elaborating why a case for "bye" is given, since the while-condition also accounts for bye? (From what I understand it's because the first input could also be bye, but this may not be self-evident)

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+
+public class Ava {
+
+    public static final String LINE_SEPARATOR = "_____________________________";
+    public static final int MAX_TASKS = 100;
+
+    public static void main(String[] args) {
+        String line;
+        Scanner in = new Scanner(System.in);
+        Task[] list = new Task[MAX_TASKS];
```
nmarwah7: Could this array be named slightly more descriptively (maybe taskList), since further in the code it's referenced simply as 
"list" which could be confusing with the command to "list"?

src/main/java/Ava.java
```diff
@@ -12,37 +11,123 @@ public static void main(String[] args) {
         Task[] list = new Task[MAX_TASKS];
         int counter = 0;
         greetings();
+
```
dylancmznus, line 14: code is done properly

nmarwah7 (commented):
Looks good to me. Clear and concise code, only few nits about variable naming under coding standards.

src/main/java/Ava.java
```diff
@@ -12,37 +11,123 @@ public static void main(String[] args) {
...
+                counter++;
+                break;
+
+            default:
+                addTask(new Task(line), list, counter);
+                counter++;
+                break;
             }
         } while (!line.equals("bye"));
+
```
dylancmznus: the code is properly done and well organized, following standard coding procedures

src/main/java/Task.java
```diff
@@ -1 +1,29 @@
-record Pair<T,U>(T t, U u) {}
\ No newline at end of file
+public class Task {
```
dylancmznus: code is clear and to the point

dylancmznus: the classes are properly named, although they could be more clearly defined for readers to minimize confusion

src/main/java/Ava.java
dylancmznus: logic is easy to follow

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+            }
+        } while (!line.equals("bye"));
+
+        goodbyes();
+    }
+
+    private static void printTaskList(Task[] list, int counter) {
+        System.out.println(LINE_SEPARATOR);
+        for (int i = 0; i < counter; i++) {
+            System.out.println((i + 1) + ": " + list[i].toString());
```
nictzl: What do you mean by this + 1? Perhaps you could use a constant to explain it

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
+import java.util.Scanner;
+
+public class Ava {
+
+    public static final String LINE_SEPARATOR = "_____________________________";
```
nictzl: I like the use of constants to reduce repeating code

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+            System.out.println("OK, I've marked this task as not done yet:\n  " + list[index]);
+            System.out.println(LINE_SEPARATOR);
+        }
+    }
+
+    private static int getTaskIndex(String toMark, Task[] list, int counter) {
+        int index = -1;
+        if (Character.isDigit(toMark.charAt(0))) {
+            index = Integer.parseInt(toMark) - 1;
+        } else {
```
nictzl: Can this be done in a better way to prevent nesting of else-for-if?

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+            case "bye":
+                // Exit condition: just exit the loop
+                break;
+
+            case "todo":
+                addTask(new Todo(line.substring(5)), list, counter);
+                counter++;
+                break;
+
+            case "deadline":
```
nictzl: I like how you used case statements to prevent multiple if else blocks and makes it neater

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+                int byIndex = line.indexOf("/by");
+                String deadline = line.substring(9, byIndex - 1);
+                String by = line.substring(byIndex + 4);
+                addTask(new Deadline(deadline, by), list, counter);
+                counter++;
+                break;
+
+            case "event":
```
manavm12: use of switch, case statements reduces nesting and improves code readability. Good job

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+                int fromIndex = line.indexOf("/from");
+                String event = line.substring(6, fromIndex - 1);
+                int toIndex = line.indexOf("/to");
```
manavm12: You can consider having different functions to process adding events, todos and deadlines into the list as main function seems too long

src/main/java/Ava.java
```diff
@@ -0,0 +1,143 @@
...
+
+        do {
+            line = in.nextLine();
+            switch (line.split(" ")[0]) {
+            case "list":
+                printTaskList(list, counter);
+                break;
+
+            case "mark":
+                handleMark(line, list, counter);
```
manavm12: Good use of functions to handle mark and unmark. Improves code readability.

src/main/java/Ava.java
```diff
@@ -0,0 +1,197 @@
...
+        System.out.println(LINE_SEPARATOR);
+    }
+
+    private static void handleMark(String line, Task[] list, int counter) throws InvalidInputException {
+<<<<<<< HEAD
+        if (line.length() == 4) {
+            throw new InvalidInputException("mark/unmark cannot be empty!");
+        }
+=======
+>>>>>>> branch-Level-5
```
TVageesan: You seem to have a broken merge conflict here, you might want to clean it up

src/main/java/Ava.java
```diff
@@ -0,0 +1,197 @@
...
+
+                case "todo":
+                    counter = addTodo(line, list, counter);
+                    break;
+
+                case "deadline":
+                    counter = addDeadline(line, list, counter);
+                    break;
+
+                case "event":
```
TVageesan: The strings used for comparison here are magic strings. Consider using a enum or constants to represent each command word instead.

src/main/java/Ava.java
```diff
@@ -0,0 +1,197 @@
...
+        } else {
+            for (int i = 0; i < counter; i++) {
+                if (list[i].isTask(toMark)) {
+                    index = i;
+                    break;
+                }
+            }
+        }
+        return index;
+    }
```
TVageesan: I like the initiative to add the extra functionality of matching tasks both by index and by description!

src/main/java/Ava.java
```diff
@@ -0,0 +1,197 @@
...
+        if (fromIndex == -1) {
+            throw new InvalidEventException("Please use format: event [description] /from [start] /to [end]");
+        }
+        String event = line.substring(5, fromIndex - 1).trim();
+        int toIndex = line.indexOf("/to");
+        if (toIndex == -1) {
+            throw new InvalidEventException("Please use format: event [description] /from [start] /to [end]");
+        }
+        String from = line.substring(fromIndex + 5, toIndex - 1).trim();
+        String to = line.substring(toIndex + 3).trim();
```
TVageesan: Consider using .split() to simplify the logic here, or using constants to wrap the magic numbers '3','5', etc. This general principle can apply to any string processing logic you have throughout your code.

PR #135, opened 2025-02-04
[ivan-tan] iP

ivan-tan:


src/main/java/Ashley.java
```diff
@@ -0,0 +1,42 @@
+import java.util.Arrays;
+import java.util.Scanner;
+
+public class Ashley {
+    public static final String LINE_SEPERATOR = "---------------------------------------------------";
```
vegetablestabber, line 5: Nit-pick but the spelling of `SEPARATOR` is incorrect.

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
...
+        return taskId;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public void setDone(){
+        isDone = true;
+    }
```
vegetablestabber, line 22: Looks great! Keep in mind you will need a method to set your task as incomplete as well.

src/main/java/Ashley.java
```diff
@@ -0,0 +1,42 @@
...
+        System.out.println("Hello! I'm Ashley");
+        System.out.println("What can I do for you?");
+        System.out.println(LINE_SEPERATOR);
+
+        String line;
+        Scanner in = new Scanner(System.in);
+        line = in.nextLine();
+        TaskManager taskManager = new TaskManager();
+
+        while (line != null) {
```
vegetablestabber, line 17: Maybe you consider using a `switch` statement instead? This might be helpful as you're only checking for the equivalence for `line` with a String literal, which a `switch` statement does implicitly. However, an `if`-`else` statements does allow for more specificity which you may use down the line.

vegetablestabber (commented):
Looks great!

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
+    private String description;
+    private boolean isDone;
```
shuu4, line 3: Good boolean naming method that starts with prefix 'is'!

src/main/java/Ashley.java
```diff
@@ -0,0 +1,42 @@
...
+                        taskManager.setAsDone(taskId);
+                        System.out.println("Task " + taskId + " has been marked as done.");
+                        break; // Exit loop as soon as a digit is found
+                    }
+                }
+            }
+            else if (line.equals("list")) {
+                taskManager.listTasks();
+            } else {
+                taskManager.addTask(line);
```
shuu4, line 35: I like your namings! All are camelCase following the standard naming conventions.

src/main/java/Ashley.java
```diff
@@ -0,0 +1,42 @@
+import java.util.Arrays;
+import java.util.Scanner;
+
+public class Ashley {
+    public static final String LINE_SEPERATOR = "---------------------------------------------------";
+
+    public static void main(String[] args) {
+        System.out.println("Hello! I'm Ashley");
+        System.out.println("What can I do for you?");
+        System.out.println(LINE_SEPERATOR);
```
shuu4, line 10: Don't forget this typo as well^^

src/main/java/Ashley.java
```diff
@@ -0,0 +1,42 @@
...
+                break;
+            }
+            else if (line.contains("mark")) {
+                for (char c : line.toCharArray()) {
+                    if (Character.isDigit(c)) {
+                        int taskId = Integer.parseInt(Character.toString(c));
+                        taskManager.setAsDone(taskId);
+                        System.out.println("Task " + taskId + " has been marked as done.");
+                        break; // Exit loop as soon as a digit is found
+                    }
```
shuu4, line 29: Would it be better if using split() and Integer.parseInt() instead of looping through every character?

PR #136, opened 2025-02-04
[Chua Feng Yuan] iP

srfl0:


src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+                space = (word.equals(line[line.length - 1])) ? "" : " ";
+                System.out.print(space);
+            }
+            System.out.println();
+            System.out.println(separator);
+        }
+
+    }
+
+    public static void add() {
```
szeyingg: method name could be more specific

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+        System.out.println("I'll now take a note on what you typed. Type 'bye' by itself to make me stop.");
+
+        String[] line;
+        Scanner addScan;
+
+        String[] store = new String[50]; //limit list to store only 50 items
+        int storeIndex = 0;
+        String newline;
```
szeyingg: variable name should be "newLine" to follow coding standards

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+            addScan = new Scanner(System.in);
+            System.out.print("You said: ");
+            newline = addScan.nextLine();
+            line = newline.split(" ");
+
+            switch (line[0]) {
+            case "bye":
+                return;
+
+            case "list":
```
szeyingg: remove extra line in between cases for to follow standard layout

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+                space = (word.equals(line[line.length - 1])) ? "" : " ";
+                System.out.print(space);
+            }
+            store[storeIndex] = newline;
+            storeIndex++;
+            System.out.println();
+            System.out.println(separator);
+        }
+
+    }
```
szeyingg: can remove extra space between brackets

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+
+    public static void tasklist() {
```
szeyingg: function name should be properly capitalised, change to "taskList"

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+                space = (word.equals(line[line.length - 1])) ? "" : " ";
+                System.out.print(space);
+            }
+            store[storeIndex] = newTask;
+            storeIndex++;
+            System.out.println();
+            System.out.println(separator);
+        }
+
+    }
```
szeyingg: extra space

szeyingg (commented):
code mostly follows coding standards, but could be more mindful about redundant spacings between lines of code as it sometimes affects readability.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+
+            case "add":
+            case "2":
+                add();
+                break;
+
+            case "tasking":
+            case "3":
+                tasklist();
+                break;
```
Lukapeng77: For the case types, instead of using the magic numbers "1", "2", "3", can just keep like the "add", "echo", 'list" which can be more meaningful.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+                    System.out.println("This task was not done yet.");
+                }
+                else {
+                    System.out.println("Okay, I'll mark this task as not done");
+                    itemToUnmark.isDone = false;
+                    System.out.println("[" + itemToUnmark.getStatusIcon() + "] " + itemToUnmark.description);
+                }
+                System.out.println(separator);
+                continue;
+            }
```
Lukapeng77: Consider resue the same part of code snippets for the mark and unmark parts to avoid unnecessary redundancy and make use of the codes.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+                }
+                continue;
+            }
+
+            System.out.println();
+
+            System.out.print("added: ");
+            for (String word : line) {
+                System.out.print(word);
+                space = (word.equals(line[line.length - 1])) ? "" : " ";
```
Lukapeng77: It's better to create a new variable to store the line.length -1 value and better explain the variable meaning, avoid using magic numbers.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,221 @@
...
+
+    public static void echo() {
+        System.out.println("I'll now repeat after you! Type 'bye' by itself to make me stop.");
+
+        String[] line;
+        Scanner echoScan;
+
+        String space; //variable to determine if spacing is needed for purposes of echoing
+
+        while (true) {
```
Lukapeng77: For check the continue condition, can create a new variable to store and explain the meaning of the variable.
Eg: Boolean isExist = false; To check for the continue condition.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,151 @@
+import java.util.Scanner;
+
+
+public class Ixo {
+
+    public final static String SEPARATOR = "____________________________________________________";
+
+    public static void menu() {
+        System.out.println("1. Task List");
+    }
```
sevenseasofbri: Function names should ideally be verbs. So a name like `showMenu` or `printMenu` would be more apt.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public static void taskList() {
```
sevenseasofbri: Similarly for this, function names should be verbs. Also, the function name does not appear to match what is happening inside it. Additionally, thif function is >30LOC. Consider breaking it up into multiple smaller functions.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,151 @@
...
+
+        Scanner lineScan;
+        Task[] taskStore = new Task[100]; //limit list to store only 100 items
```
sevenseasofbri: Try and avoid the usage of magic literals by storing them in variables or constants. Something like `MAX_TASKS` or something would be better.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,151 @@
...
+                        System.out.println("Well done! task " + taskStore[taskToMark] + " has been marked.");
+                    } else {
+                        System.out.println("The task " + taskStore[taskToMark] + " has been unmarked.");
+                    }
+                    taskStore[taskToMark].isDone = isMark;
+                } catch (NumberFormatException e) { //for non-number input
+                    System.out.println("You did not enter a number");
+                } catch (ArrayIndexOutOfBoundsException e) { //for out of bounds index entered
+                    System.out.println("You did not enter a valid corresponding number");
+                }
```
sevenseasofbri: This is deep nesting and can borderlines arrowhead code. Refactor the internal indents to another function.

src/main/java/Ixo.java
```diff
@@ -0,0 +1,151 @@
...
+            default:
+                lineScan.nextLine();
+                System.out.println("Unknown command. Type 'bye' to exit");
+                break;
+            }
+        }
+    }
+
+
+    public static void command() {
```
sevenseasofbri: Again, the name of this function is not apt. It needs to be a verb. 
```suggestion
    public static void parseCommand() {
```

sevenseasofbri (commented):
It appears as though level 6 and 7 have not been completed yet? Please complete it soon. Do look into the java coding standard, because the naming for the functions needs to be improved.

PR #141, opened 2025-02-04
[lakshm1i] iP

lakshm1i:


src/main/java/Yale.java
```diff
@@ -0,0 +1,51 @@
+import java.util.Scanner;
+
+public class Yale {
+    private static String[] tasks = new String[100]; //Array to store tasks
```
terryasdf: You could use a constant to store the 100 value.

terryasdf (commented):
Well done!

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+    protected String description;
+    protected boolean isDone;
+
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
```
Alfred-Goh02, line 26: use K&R style brackets for if else statement

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,51 @@
+public class TaskManager {
+
+    private static final int MAX_TASKS = 100;
+    private final Task[] taskList = new Task[MAX_TASKS]; //Array to store tasks
+    private int taskCounter = 0; //Track number of tasks stored
+
```
Alfred-Goh02: Consider separating printing statements and adding it into your UI class.

src/main/java/Yale.java
```diff
@@ -0,0 +1,88 @@
...
+            } catch (YaleException e) {
+                System.out.println("Sorry! " + e.getMessage());
+            }
+        }
+    }
+
+    private static void processUserInput(String userInput, TaskManager taskManager) throws YaleException {
+        if (userInput.equalsIgnoreCase("list")) {
+            taskManager.displayTaskList();
+        } else if (userInput.startsWith("mark ")) {
```
Alfred-Goh02: All these command strings like list, mark etc. should be declared as a constant at the start of the file. Avoid magic literals.

PR #148, opened 2025-02-04
[szeyingg] iP

szeyingg:


src/main/java/Mira.java
```diff
@@ -0,0 +1,127 @@
...
+    public static void addEvent(String command) {
+        int fromIndex = command.indexOf(" /from ");
+        int toIndex = command.indexOf(" /to ");
+
+        if (fromIndex == -1 || toIndex == -1 || fromIndex > toIndex) {
+            System.out.println("invalid");
+        } else {
+            String description = command.substring(0, fromIndex);
+            String from = command.substring(fromIndex + 7, toIndex);
+            String to = command.substring(toIndex + 5);
```
Flaaaash: Instead of using index to keep track of characters in a string, try to use .split to separate content of a string as it avoids creating magic numbers like 5 and 7.

src/main/java/Mira.java
```diff
@@ -0,0 +1,127 @@
...
+            String description = command.substring(0, byIndex);
+            String by = command.substring(byIndex + 5);
+
+            list[listLength] = new Deadline(description, by);
+            reportTaskAdded(list[listLength]);
+            listLength++;
+        }
+    }
+
+    public static void addEvent(String command) {
```
Flaaaash: Instead of having an add... method for each sub-class of a task, you may consider creating an addTask(Task task) method where you instantiate the tasks as the input parameters of the method. This way you can avoid creating a add... method every time you create another sub-class of a task, and therefore improve readability and make your code easier to maintain.

Flaaaash (commented):
Generally good coding practices.
Some changes can be done to make the code easier to maintain.

src/main/java/Mira.java
```diff
@@ -0,0 +1,127 @@
...
+        t.markAsDone();
+        System.out.println("Presto, you did it! Task " + idx + " has been conquered!");
+    }
+
+    public static void unmarkTask(String num) {
+        int idx = Integer.parseInt(num);
+        Task t = list[idx -1];
+        t.markAsUndone();
+        System.out.println("Tough luck! Task " + idx + " has come back stronger!");
+    }
```
HightechTR, line 22: A lot of these methods can be put into Task.java or Todo.java to make Mira.java more readable

src/main/java/Mira.java
```diff
@@ -0,0 +1,127 @@
...
+            switch (action) {
+                case "bye": exit(); break;
+                case "list": printList(); break;
+                case "todo": addTodo(param); break;
+                case "deadline": addDeadline(param); break;
+                case "event": addEvent(param); break;
+                case "mark": markTask(param); break;
+                case "unmark": unmarkTask(param); break;
+                default: System.out.println("Oh no... seems like your spell " + action + " failed, try another?");
+            }
```
HightechTR: This switch statement doesn't follow standards, you should follow the standard:
`switch (action) {`
`case "bye":`
`    exit();`
`    break;`

or use lambda style:
`switch (action) {`
`    case "bye" -> exit();`

src/main/java/Mira.java
```diff
@@ -0,0 +1,127 @@
+import java.util.Arrays;
+import java.util.Scanner;
+
+public class Mira {
+    private static Task[] list;
```
HightechTR: Standards: It would be better to name this Task array `tasks`

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+        return "[" + getStatusIcon() + "] " + description;
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsUndone() {
+        this.isDone = false;
+    }
```
HightechTR: Would help to have a single setter `setDone(boolean isDone)` here

src/main/java/Mira.java
```diff
@@ -0,0 +1,48 @@
...
+                switch (action) {
+                    case "bye" -> exit();
+                    case "list" -> Task.printList();
+                    case "todo" -> Todo.addTodo(param);
+                    case "deadline" -> Deadline.addDeadline(param);
+                    case "event" -> Event.addEvent(param);
+                    case "mark" -> Task.markTask(param);
+                    case "unmark" -> Task.unmarkTask(param);
+                    case "help" -> System.out.println(Text.COMMAND_LIST);
+                    default -> throw new InvalidCommandException();
```
bljhty: You may want to follow the switch-case format given

src/main/java/mira/task/Task.java
```diff
@@ -0,0 +1,92 @@
...
+        try {
+            int idx = Integer.parseInt(num);
+            if (idx <= 0 || idx > listLength) {
+                throw new NumberFormatException();
+            }
+            Task t = tasks[idx - 1];
+            t.setDone(isDone);
+
+            String actionText = isDone ? "conquered" : "come back stronger";
+            System.out.print(isDone ? "Presto, you did it! " : "Tough luck! ");
```
bljhty: The use of magic literals is not recommended

PR #152, opened 2025-02-05
[Anak Agung Gde Yogi Pramana] iP

IgoyAI:


src/main/java/Asep.java
```diff
@@ -0,0 +1,187 @@
...
+public class Asep {
+    private static final int MAX_TASKS = 100;
+    private final Task[] tasks;
+    private int taskCount;
+
+    /**
+     * Constructs a new Asep chatbot.
+     */
+    public Asep() {
+        tasks = new Task[MAX_TASKS];
```
Carlton369: Good use of constants to improve readability by avoiding "magic numbers"

src/main/java/Asep.java
```diff
@@ -0,0 +1,187 @@
...
+                processEventCommand(userInput);
+            } else {
+                System.out.println("Unknown command.");
+            }
+        }
+
+        scanner.close();
+    }
+
+    private void printGreeting() {
```
Carlton369: Clear naming convention for your methods and variables, makes it easy to understand

src/main/java/Asep.java
```diff
@@ -0,0 +1,187 @@
...
+        System.out.println("____________________________________________________________");
+        System.out.println(" Hello! I'm Asep");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+    }
+
+    private void printFarewell() {
+        System.out.println("____________________________________________________________");
```
Carlton369: This border could be changed into a const or a separate method to make the code neater

src/main/java/Task.java
```diff
@@ -0,0 +1,53 @@
...
+    protected String getExtraInfo() {
+        return "";
+    }
+
+    /**
+     * Returns the string representation of the task.
+     * Example: [T][ ] borrow book
+     *
+     * @return the formatted task string
+     */
```
Carlton369: Good use of comments to make it clear what the functions are doing

Carlton369 (commented):
LGTM! Your code is very readable and clear!

src/main/java/Asep.java
chuanhuiw: I agree with this! This helps improve your code quality without introducing unknown numbers

src/main/java/Asep.java
```diff
@@ -0,0 +1,187 @@
...
+            } else if (userInput.toLowerCase().startsWith("todo ")) {
+                processTodoCommand(userInput);
+            } else if (userInput.toLowerCase().startsWith("deadline ")) {
+                processDeadlineCommand(userInput);
+            } else if (userInput.toLowerCase().startsWith("event ")) {
+                processEventCommand(userInput);
+            } else {
+                System.out.println("Unknown command.");
+            }
+        }
```
chuanhuiw: very neat loop without nested loops, nice!

src/main/java/Task.java
chuanhuiw: very good documentation done

chuanhuiw (commented):
Overall code is clear and concise, good job!

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+        taskCount = 0;
+        tasks = new ArrayList<>();
+        storage = new Storage("./data/asep.txt");
+        tasks = storage.loadTasks();
+    }
+
+    /**
+     * Starts the chatbot.
+     */
+    public void run() {
```
sevenseasofbri, line 25: This method is a little over 30 lines long, which may hinder readability. Consider refactoring the deeply nested parts of the code into another function.

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+                    processUnmarkCommand(userInput);
+                } else if (userInput.toLowerCase().startsWith("todo ")) {
+                    processTodoCommand(userInput);
+                } else if (userInput.toLowerCase().startsWith("deadline ")) {
+                    processDeadlineCommand(userInput);
+                } else if (userInput.toLowerCase().startsWith("event ")) {
+                    processEventCommand(userInput);
+                } else {
+                    throw new AsepException("Oops! I don't recognize this command. Please try again.");
+                }
```
sevenseasofbri: This code is hitting the 3rd level of indentation, it may help to refactor the code within this try block to another function.

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+            } catch (AsepException e) {
+                System.out.println("____________________________________________________________");
+                System.out.println(" " + e.getMessage());
+                System.out.println("____________________________________________________________");
+            } catch (Exception e) {
+                System.out.println("____________________________________________________________");
+                System.out.println(" An unexpected error occurred: " + e.getMessage());
+                System.out.println("____________________________________________________________");
```
sevenseasofbri: I notice you have other functions to print greeting and farewell. Maybe you could make a function like `printError` and pass the `e` variable to that to print the errors. This would allow you to maintain the same level of abstraction within this method.

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+        System.out.println(" Here are the tasks in your list:");
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println(" " + (i + 1) + "." + tasks[i].toString());
+        }
+        System.out.println("____________________________________________________________");
+    }
+
+    private void processMarkCommand(String command) {
+        String[] parts = command.split("\\s+");
+        if (parts.length != 2) {
```
sevenseasofbri: Good use of "[keeping the happy path prominent"](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent)!

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+        System.out.println("____________________________________________________________");
+        System.out.println(" Here are the tasks in your list:");
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println(" " + (i + 1) + "." + tasks[i].toString());
+        }
+        System.out.println("____________________________________________________________");
+    }
+
+    private void processMarkCommand(String command) {
+        String[] parts = command.split("\\s+");
```
sevenseasofbri: `parts` may not be such an intuitive name foe this variable. Consider naming it something along the lines of `commandParts` or `delimitedParts`

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    private void processUnmarkCommand(String command) {
+        String[] parts = command.split("\\s+");
+        if (parts.length != 2) {
+            System.out.println("Invalid command format. Usage: unmark <task number>");
+            return;
+        }
+        try {
+            int index = Integer.parseInt(parts[1]) - 1;
```
sevenseasofbri: Avoid using magic numbers, however, if it is not possible in this case, it would be better to have a concise comment on what is being done here.

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+            throw new AsepException("Deadline description and time cannot be empty.");
+        }
+
+        addTask(new Deadline(description, by));
+    }
+
+    private void processEventCommand(String command) throws AsepException {
+        String content = command.substring(6).trim();
+        int fromIndex = content.indexOf("/from");
+        int toIndex = content.indexOf("/to");
```
sevenseasofbri: These substrings can be stored in consts to avoid the usage of magic numbers.

src/main/java/Asep.java
```diff
@@ -0,0 +1,232 @@
...
+            throw new AsepException("Invalid task number.");
+        }
+    }
+
+    public static void main(String[] args) {
+        new Asep().run();
+    }
+
+
+
```
sevenseasofbri: You might want to remove this unnecessary spacing

src/main/java/Storage.java
```diff
@@ -0,0 +1,74 @@
+import java.io.*;
```
sevenseasofbri: Wildcard imports should be avoided. Import the classes you need only.

src/main/java/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                Task task = parseTask(line);
+                if (task != null) {
+                    tasks.add(task);
+                }
+            }
```
sevenseasofbri: Prevent deep nesting by refactoring deep code into a different method.

sevenseasofbri (commented):
Good work so far! The usage of magic numbers is quite liberal in your code so I would suggest you to use constants/variables instead where necessary (and possible). This will definitely improve readability.

PR #153, opened 2025-02-05
[Tang Chi Yan] iP

Lyam-T:


src/main/java/Jake.java
```diff
@@ -0,0 +1,121 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+import java.util.regex.Pattern;
+
+public class Jake {
+    private static final TaskPool taskPool = new TaskPool();
+    private static final String lineSeparator = "  -----------------------------------------------------------------";
```
edwardrl101: I like how you extracted out a constant for the separator to prevent it from being a magic literal!

src/main/java/Jake.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+        System.out.println(lineSeparator);
+    }
+
+    public static void main(String[] args) {
+        printWelcomingMsg();
+        readAndHandleInput();
+        printByeMsg();
+    }
```
edwardrl101: Good job structuring the code out logically!

src/main/java/TaskPool.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public void addEvent(String name, String from, String to) {
+        tasks.add(new Event(name, from, to));
+    }
+
+    /**
+     * Marks a task as done.
+     * @param taskNumber the task number as shown in the list command, starting from 1.
+     */
+    public void markDone(int taskNumber) {
```
edwardrl101: Perhaps a more intuitive method name such as markTaskAsDone can be used since the Task class has the same method name?

src/main/java/TaskPool.java
```diff
@@ -0,0 +1,52 @@
...
+        tasks.get(taskNumber - 1).markDone();
+    }
+
+    /**
+     * Marks a task as undone.
+     * @param taskNumber the task number as shown in the list command, starting from 1.
+     */
+    public void markUndone(int taskNumber) {
```
edwardrl101: Perhaps a more intuitive method name such as markTaskAsUndone can be used since the Task class has the same method name?

src/main/java/Jake.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        return result.toArray(new String[0]);
+    }
+
+    /**
+     * Parse the input and execute the command.
+     * @param input the input string from the user.
+     */
+    private static void handleInput(String input) {
```
edwardrl101: I like how for each case in the switch condition, the code inside it has similar levels of abstraction!

edwardrl101: Good job ensuring the code is written out at the highest level of abstraction as possible!

src/main/java/Jake.java
```diff
@@ -0,0 +1,121 @@
...
+            input = in.nextLine();
+        }
+    }
+
+    /**
+     * Parse the input string into command and arguments.
+     * @param input the input string from the user.
+     * @return an array of strings containing the command and arguments.
+     */
+    private static String[] parseInput(String input) {
```
edwardrl101: I like how you followed the camelCase and verb naming conventions for methods! Similarly for all your other methods as well!

src/main/java/Jake.java
```diff
@@ -0,0 +1,121 @@
...
+            handleInput(input);
+            input = in.nextLine();
+        }
+    }
+
+    /**
+     * Parse the input string into command and arguments.
+     * @param input the input string from the user.
+     * @return an array of strings containing the command and arguments.
+     */
```
edwardrl101: The JavaDoc documentation for the code is good, and it is able to explain the method well to the reader

src/main/java/Jake.java
```diff
@@ -0,0 +1,121 @@
...
+        taskPool.printTask(taskPool.getTaskCount());
+    }
+
+    private static void printUnmarkTaskMsg() {
+        System.out.println("  Nice! I've unmarked this task:");
+        System.out.print("    ");
+        taskPool.printTask(taskPool.getTaskCount());
+    }
+
+    private static void readAndHandleInput() {
```
edwardrl101: The naming of the method is good as it explains what it does to the reader. Similarly for most of your other methods as well

src/main/java/TaskPool.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public void addEvent(String name, String from, String to) {
+        tasks.add(new Event(name, from, to));
+    }
+
+    /**
+     * Marks a task as done.
+     * @param taskNumber the task number as shown in the list command, starting from 1.
+     */
```
edwardrl101: Perhaps this JavaDoc documentation is not needed as what the method does is already obvious?

src/main/java/TaskPool.java
Lyam-T: you are correct, thanks for the reminder

src/main/java/TaskPool.java
Lyam-T: just to make sure that reader won't input 0 as the task number

src/main/java/Jake/Jake.java
```diff
@@ -0,0 +1,199 @@
...
+     * @param input the input string from the user.
+     */
+    private static void handleInput(String input) {
+        String[] commandAndArgs = parseInput(input);
+
+        System.out.println(Message.LINE_SEPARATOR);
+        try {
+            switch (Command.valueOf(commandAndArgs[0].toUpperCase())) {
+                case LIST -> {
+                    handleList();
```
thomasjlalba: Good job on abstracting to smaller methods!

src/main/java/Jake/Message.java
```diff
@@ -0,0 +1,44 @@
+package Jake;
+
+public class Message {
+    static public final String INDENT = "  ";
```
thomasjlalba: Good job on avoiding magic literals!

src/main/java/Jake/Jake.java
```diff
@@ -0,0 +1,199 @@
...
+           System.out.println(commandAndArgs[0] + Message.INVALID_PARAMS);
+       } catch (IndexOutOfBoundsException e) {
+           System.out.println(commandAndArgs[0] + Message.INVALID_PARAMS + "\n" + Message.CHECK_CUR_TASKS);
+           taskPool.printTasks();
+       }
+    }
+
+    private static void handleEvent(String[] commandAndArgs) {
+        try {
+            taskPool.addEvent(commandAndArgs[1], commandAndArgs[2], commandAndArgs[3]);
```
thomasjlalba: This is just a personal preference, but it might be more understandable if you refactor the parameters (`commandAndArgs[0]`, etc.) to indicate what they are intended for (`name`, `from`, etc.)

thomasjlalba (commented):
Well done on your iP thus far! Do not have any complains or even nitpicks on your coding standards and quality!

PR #161, opened 2025-02-05
[Carlton369] iP

Carlton369:


src/main/java/Task/Event.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public String getFrom(){
+        return from;
+    }
+
+    public String getTo(){
+        return to;
+    }
+
+    @Override
```
chuanhuiw: I like how you utilized coding standards learnt in lecture like @Override in your ip and coding style is very neat! Wow good job!

src/main/java/Task/Task.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public void unmarkAsDone() {
+        isDone = false;
+    }
+
+    public String toString(){
+        return "[" + getStatusIcon() + "] " + description;
+    }
+}
```
chuanhuiw: I like how your indentation was done, no nested loops

src/main/java/Snorlax.java
```diff
@@ -76,73 +61,45 @@ public static void main(String[] args) {
...
+                    if (line.isEmpty()) {
+                        printErrorMsg("no input");
+                    } else if (count < 100) {
+                        printBorder();
+                        System.out.println("     added: " + line + ".....");
+                        printBorder();
+                        list[count] = new Task(line);
+                        count += 1;
+                    } else {
+                        printErrorMsg("full list");
```
chuanhuiw: good use of error messages for debugging

src/main/java/Snorlax.java
```diff
@@ -1,38 +1,22 @@
 import java.util.Scanner;
+import Task.Task;
+import Task.Deadline;
 
 public class Snorlax {
+    public static void printBorder(){
+        System.out.println("     zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz");
```
chuanhuiw: very fitting theme

src/main/java/Snorlax.java
```diff
@@ -0,0 +1,105 @@
...
+        case "completed task":
+            System.out.println("     Task is already done.....");
+            break;
+        }
+        printBorder();
+    }
+
+    public static void main(String[] args) {
+        String name = "Snorlax";
+        printIntro(name);
```
Jonathan2745: Great Job! Methods are labelled with clear names

src/main/java/Snorlax.java
```diff
@@ -0,0 +1,105 @@
...
+        case "full list":
+            System.out.println("     Sorry.... list is full.....");
+            break;
+        case "incomplete task":
+            System.out.println("     Task is not done.....");
+            break;
+        case "completed task":
+            System.out.println("     Task is already done.....");
+            break;
+        }
```
Jonathan2745: Good job separating the error cases! I think that you are missing the default branch for case though.

src/main/java/Snorlax.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public static void main(String[] args) {
+        String name = "Snorlax";
+        printIntro(name);
+
+        String line;
+        Scanner in = new Scanner(System.in);
+
+        line = in.nextLine().trim();
```
Jonathan2745: in terms of using names to explain, I feel that "line" could be improved to something like "userInput"

src/main/java/Snorlax.java
```diff
@@ -0,0 +1,105 @@
...
+                case "unmark" -> {
+                    if (splitLine.length < 2 || !splitLine[1].matches("\\d+")) {
+                        printErrorMsg("invalid task");
+                    } else {
+                        int number = Integer.parseInt(splitLine[1]) - 1;
+                        Commands.Unmark(list, number, count);
+                    }
+                }
+
+                default -> {
```
Jonathan2745: Great job including the default case here !

src/main/java/Snorlax.java
```diff
@@ -0,0 +1,105 @@
...
+        printIntro(name);
+
+        String line;
+        Scanner in = new Scanner(System.in);
+
+        line = in.nextLine().trim();
+
+        Task[] list = new Task[100];
+
+        int count = 0;
```
Jonathan2745: Similar to the above comment, I think that the name "count" could be improved to represent the number of tasks better
!

src/main/java/Commands.java
```diff
@@ -0,0 +1,46 @@
...
+            System.out.println("     Here are the tasks in the list:");
+            for (int i = 0; i < count; i++) {
+                System.out.println("     " + (i + 1) + ":["
+                        + list[i].getStatusIcon() + "] " + list[i].description + ".....");
+            }
+        }
+        Snorlax.printBorder();
+    }
+
+    public static void Mark(Task[] list, int number, int count){
```
Jonathan2745: Excellent abstraction of these actions, the indentation of the if, else if and else statements are perfect as well !

src/main/java/Snorlax.java
Carlton369: Thanks for your comment! I did not include a default case as the parameter for the error is hard coded based, hence there the default branch would never be entered. However, I do agree that it would be safer for me to include it in case I accidentally called an error that was not explicitly included here. Will add a default case in a future commit!

src/main/java/Snorlax.java
Carlton369: Thank you for your suggestion! I agree that does make it more readable, will be implemented in a future commit!

Carlton369 (commented):
Responded to comments by reviewers, will take their suggestions into account and to be fixed in future commit

src/main/java/Snorlax/CommandPackage/Commands.java
```diff
@@ -0,0 +1,240 @@
...
+    /**
+     * Prints out all items in the list
+     * If empty list, prints error message indicating empty list
+     *
+     * @param list list to be printed out
+     */
+    public static void List(ArrayList<Task> list) {
+        UI.printBorder();
+        if (list.isEmpty()) {
+            System.out.println("     No tasks yet.....");
```
yijiano, line 27: Since you are printing tab spaces very often, perhaps you can have a helper function for this? Maybe something like `printIndented` or something similar.

src/main/java/Snorlax/ParserPackage/IO.java
```diff
@@ -0,0 +1,145 @@
...
+     * @throws FullListException       If the task list is full (100 tasks limit).
+     * @throws NoInputException        If no input is provided.
+     * @throws InvalidCommandException If an unrecognized command is entered.
+     */
+    public static void handleInput (ArrayList<Task> list, String[] splitLine)  {
+        switch (splitLine[0].toLowerCase()) {
+        case "list" -> {
+            Commands.List(list);
+        }
+        case "mark" -> {
```
yijiano, line 59: Maybe you can refactor this part and clearly define what `splitLine[i]` is using variables, so that its easier to know what's going on.

src/main/java/Snorlax/ParserPackage/IO.java
```diff
@@ -0,0 +1,145 @@
...
+            } else {
+                int taskNumber = Integer.parseInt(splitLine[1]) - 1;
+                Commands.Delete(list, taskNumber);
+                FileIO.saveTasks(list);
+            }
+        }
+        case "deadline" -> {
+            if (splitLine.length < 2) {
+                throw new InvalidTaskException();
+            } else if (list.size() >= 100) {
```
yijiano, line 89: Is there a need to use this magic number?

yijiano (commented):
Overall good effort. Try to reduce redundant code across the board.

PR #162, opened 2025-02-05
[Jackson Hu] iP

jhu0748:


src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
+import java.util.Scanner;
+
+public class Nova {
+    public static void main(String[] args) {
+        String logo = "Nova";
```
xhuinn: Instead of naming String as "Nova", perhaps it would be better to rename it as "NOVA" as it is a constant String

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
...
+        printLine();
+        System.out.println("     Hello! I'm " + logo);
+        System.out.println("     What can I do for you?");
```
xhuinn: Instead of manually inputing spaces, perhaps it would be better to use the tab escape sequence (\t)

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
...
+                break;
+            }
+        }
+
+        scanner.close();
+    }
+
+    // helper method to print lines separating user input and command output
+    private static void printLine() {
+        System.out.println("    ____________________________________________________________");
```
xhuinn: Instead of manually printing the divider out, perhaps it would be better to change it to a String constant

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
...
+                    printLine();
+                }
+                break;
+            }
+            default:
+                tasks[numTasks] = input;
+                numTasks++;
+                printLine();
+                System.out.println("     added: " + input);
+                printLine();
```
xhuinn: Instead of using printLine(); to add a new line, perhaps it would be better to use the newline escape sequence (\n)

xhuinn (commented):
Generally looks good.
Just a few nits to fix.

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
+import java.util.Scanner;
+
+public class Nova {
+    public static void main(String[] args) {
+        String logo = "Nova";
+        printLine();
+        System.out.println("     Hello! I'm " + logo);
```
cj143n: Might not need the logo variable and just type in "Hello! I'm Nova" instead

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
...
+        System.out.println("     What can I do for you?");
+        printLine();
+        Scanner scanner = new Scanner(System.in);
+        String[] tasks = new String[100];
```
cj143n: May be better to name it taskList signify that it is a list more literally

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
...
+                    System.out.println("       [X] " + tasks[taskNum]);
+                    printLine();
+                } else {
+                    printLine();
+                    System.out.println("     Invalid task number.");
+                    printLine();
+                }
+                break;
+            }
+            case "unmark": { // unmark task
```
cj143n: May be better to extract them into functions to be neater

src/main/java/Nova.java
```diff
@@ -0,0 +1,80 @@
...
+                System.out.println("     added: " + input);
+                printLine();
+                break;
+            }
+        }
+
+        scanner.close();
+    }
+
+    // helper method to print lines separating user input and command output
```
cj143n: I like the comment explaining the function's usage

cj143n (commented):
Overall clean and clear code :)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+    private String by;
+
+    public Deadline(String description, String by) throws NovaException{
+        super(description.trim());
+        if (description.trim().isEmpty()) {
+            throw new NovaException("OOPS!!! Sorry, but the description of a deadline cannot be empty.");
+        }
+//        if(description.charAt(0) != ' ') {
+//            throw new NovaException(("OOPS!! Sorry, but that is an unrecognized command."));
+//        }
```
wenxin-c: Please remove commented code.

src/main/java/Event.java
```diff
@@ -0,0 +1,32 @@
...
+        }
+        if (from == null || from.trim().isEmpty()) {
+            throw new NovaException("OOPS!!! Sorry, but the event must have a valid /from date.");
+        }
+        if (to == null || to.trim().isEmpty()) {
+            throw new NovaException("OOPS!!! Sorry, but the event must have a valid /to date.");
+        }
+//        if(description.charAt(0) != ' ') {
+//            throw new NovaException(("OOPS!! Sorry, but that is an unrecognized command."));
+//        }
```
wenxin-c: Refer to the previous comment on commentd code.

src/main/java/Nova.java
```diff
@@ -0,0 +1,74 @@
...
+                    try {
+                        String[] parts = input.substring(6).split(" /from | /to ");
+                        if (parts.length < 3) {
+                            throw new NovaException("OOPS!! An event must have a description, /from, and /to times.");
+                        }
+                        taskList.addTask(new Event(parts[0], parts[1], parts[2]));
+                    } catch (NovaException e) {
+                        Ui.printSeparatorLine();
+                        System.out.println("\t " + e.getMessage());
+                        Ui.printSeparatorLine();
```
wenxin-c: You could consider abstracting add deadline and add event into methods as well, and split ```addTask``` into more specific kinds of tasks.

src/main/java/Nova.java
```diff
@@ -0,0 +1,74 @@
...
+            try {
+                String input = scanner.nextLine();
+                if (input.equals("bye")) {
+                    Ui.printSeparatorLine();
+                    System.out.println("\t Bye. Hope to see you again soon!");
+                    Ui.printSeparatorLine();
+                    break;
+                } else if (input.equals("list")) {
+                    taskList.listTasks();
+                } else if (input.startsWith("todo ")) {
```
wenxin-c: One suggestion is that you could filter out the command word and use ```switch``` and ```case``` to check for command, instead of using ```startsWith()``` multiple times.

src/main/java/Nova.java
```diff
@@ -0,0 +1,74 @@
...
+                    taskList.deleteTask(taskIndexDelete);
+                } else {
+                    Ui.printSeparatorLine();
+                    System.out.println("\t OOPS!! I apologize but that is an unrecognized command, please try again.");
+                    Ui.printSeparatorLine();
+                }
+            } catch(NovaException e) {
+                Ui.printSeparatorLine();
+                System.out.println("\t " + e.getMessage());
+                Ui.printSeparatorLine();
```
wenxin-c: You could consider abstracting this into a method as well since this is used quite often when catching exceptions.

src/main/java/TaskList.java
```diff
@@ -0,0 +1,148 @@
+import java.util.ArrayList;
+import java.io.*;
```
wenxin-c: Please import libraries explicitly instead of using wildcard.

src/main/java/Todo.java
```diff
@@ -0,0 +1,24 @@
...
+        super(description);
+
+        if(description.trim().isEmpty()) {
+            throw new NovaException("OOPS!! Sorry, but the todo description can't be empty. Please include a description and try again.");
+        }
+/*
+        if(description.trim().charAt(0) == ' ') { //
+            throw new NovaException(("OOPS!! Sorry, but that is an unrecognized command."));
+        }
+*/
```
wenxin-c: Refer to previous comment on commented code.

wenxin-c (commented):
Good job for IP and overall good use of OOP. You might consider abstracting methods further so that the code will be more organised and remeber to remove commented code as well.

PR #167, opened 2025-02-05
[vihaan27] iP

vihaan27:


src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+            String[] words = line.split(" ");
+            if(line.equals("list")){
+                printList(tasks, taskCount);
+            }
+            else if(words[0].equals("mark")){
+                int pos = Integer.parseInt(words[1]);
+                if (pos > 0 && pos <= taskCount) {
+                    tasks[pos - 1].setComplete(true);
+                }
+                else{
```
QuyDatNguyen: Need space between else and {

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+                    System.out.println("No item found at position "+ pos);
+                }
+            }
+            else if(words[0].equals("unmark")){
```
QuyDatNguyen: Need space before {

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+            else if (!line.equals("bye")) {
+                tasks[taskCount] = new Task(line, false);
+                taskCount++;
+                System.out.println("Added: " + line);
+            }
+        } while(!line.equals("bye"));
+
+        System.out.println(exit);
+    }
+    public static void printList(Task[] tasks, int taskCount){
```
QuyDatNguyen: Need space before {

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public void setItem(String item) {
+        this.item = item;
+    }
+
+    public boolean isComplete() {
+        return isComplete;
+    }
+
+    public void setComplete(boolean complete) {
```
QuyDatNguyen: the variable "complete" should be named sound like booleans more

QuyDatNguyen (commented):
Need to consider the spaces between { more

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+                }
+            }
+            else if(words[0].equals("unmark")){
+                int pos = Integer.parseInt(words[1]);
+                if (pos > 0 && pos <= taskCount) {
+                    tasks[pos - 1].setComplete(false);
+                }
+                else{
+                    System.out.println("No item found at position "+ pos);
+                }
```
brand0nnn: if-else statements should follow the Java coding standard,

if {
    ....
} else if () {
    ....
}

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+                System.out.print((i+1) + ". [X] ");
+            }
+            else{
+                System.out.print((i+1) + ". [ ] ");
+            }
+            System.out.println(tasks[i].getItem());
+
+        }
+    }
+}
```
brand0nnn: Neat print function!

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+        Scanner in = new Scanner(System.in);
+        String line;
+        do {
+            System.out.print("~ ");
+            line = in.nextLine();
+            String[] words = line.split(" ");
+            if(line.equals("list")){
+                printList(tasks, taskCount);
+            }
+            else if(words[0].equals("mark")){
```
brand0nnn: Do indicate a spacing for the if-else block conditions too like the following

if (condition) {
}

instead of 

if(condition){
}

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
+import java.util.Scanner;
+public class Chip {
+    public static void main(String[] args) {
+        String greeting = "Hello! I'm Chip. What can I do for you?";
+        String exit = "Bye. Hope to see you again soon!";
```
TiangSoonYong: I like the idea of defining these lines as individual string. Since it should not be modified, perhaps they can be defined as  `CONSTANT(s)`

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+        System.out.println(greeting);
+        Task[] tasks = new Task[100];
```
TiangSoonYong: The max size for tasks `100` is a case of magic literals, perhaps it can be defined as a `CONSTANT`

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+        int taskCount = 0;
+
+        Scanner in = new Scanner(System.in);
+        String line;
+        do {
+            System.out.print("~ ");
+            line = in.nextLine();
+            String[] words = line.split(" ");
```
TiangSoonYong: I find `words[i]` slightly difficult to understand. Maybe they can be separated into individual strings such as `String command = word[i]`. This may help with code clarity.

src/main/java/Chip.java
```diff
@@ -0,0 +1,58 @@
...
+                }
+                else{
+                    System.out.println("No item found at position "+ pos);
+                }
+            }
+            else if (!line.equals("bye")) {
+                tasks[taskCount] = new Task(line, false);
+                taskCount++;
+                System.out.println("Added: " + line);
+            }
```
TiangSoonYong: > if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}

The if-else block should follow this coding standard

src/main/java/Task.java
TiangSoonYong: > Boolean variables/methods should be named to sound like booleans

Perhaps `complete` can be renamed as `isComplete`

TiangSoonYong (commented):
Overall, great work on formatting outputs! The only big issue I observed is code formatting. Let the IDE handle it with `Ctrl+Alt+L` (if you are using IntelliJ). All the best!

src/main/java/Chip.java
```diff
@@ -0,0 +1,256 @@
...
+public class Chip {
+    public static final int MAX_TASKS = 100;
+    public static ArrayList<Task> tasks = new ArrayList<>();
+    public static final String DIRECTORY = "data";
+    public static final String FILE_PATH = "data/chip.txt";
+    public static int taskCount = 0;
+
+    public static void main(String[] args) {
+        String greeting = "Hi! I'm Chip. What can I do for you?";
+        String exit = "Bye. See you again soon!";
```
irving11119: These can be considered magic strings. Perhaps you could follow the format for magic values (e.g. declaring them as static, upper caps etc.)

src/main/java/Chip.java
```diff
@@ -0,0 +1,256 @@
...
+        Task taskToAdd;
+        switch (details[0]) {
+        case "T":
+            taskToAdd = new ToDo(details[2]);
+            break;
+        case "E":
+            taskToAdd = new Event(details[2], details[3], details[4]);
+            break;
+        case "D":
+            taskToAdd = new Deadline(details[2], details[3]);
```
irving11119: I believe you're missing a break here.

src/main/java/Chip.java
```diff
@@ -0,0 +1,256 @@
...
+        String[] words = input.split(" ");
+        int descriptionStart = input.indexOf(words[1]);
+        int descriptionEnd = input.indexOf("/from");
+        String description = input.substring(descriptionStart, descriptionEnd - 1);
+        int fromEnd = input.indexOf("/to");
+        String from = input.substring(descriptionEnd + 6, fromEnd);
+        String to = input.substring(fromEnd + 4);
+        Event event = new Event(description.trim(), from.trim(), to.trim());
+        addTask(event);
+    }
```
irving11119: You can try restructuring your code such that the it's more readable. You can use blank lines to separate groups of related statements which makes it easier for code reviewers to understand your logic

src/main/java/InvalidCommandException.java
```diff
@@ -0,0 +1,5 @@
+public class InvalidCommandException extends ChipException{
+    public InvalidCommandException(){
+        super("I'm sorry, I didn't understand that :(");
```
irving11119, line 3: Try to avoid magic strings as much as possible.

irving11119 (commented):
Overall, your code adheres to the coding standards and is readable. Some feedback from me would just be to structure the code so that is more readable by grouping similar statements together and separating different groups using whitespaces. Another minor nitpick (not part of the standards or quality) is that some of your return statements and can be removed without impacting the functionality of the program. This can make your program more readable.

PR #170, opened 2025-02-05
[Stevexchia] iP

Stevexchia:


src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+        this.isDone = false;
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
```
ravi-viswa105: I like how clear the naming of the methods are

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
+import java.util.Scanner;
+import java.util.ArrayList;
```
ravi-viswa105: I like that the imported classes are listed explicitly

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+                        System.out.println(" Invalid task number.");
+                    }
+                } catch (NumberFormatException e) {
+                    System.out.println(" Please enter a valid number.");
+                }
+            } else {
+                Task newTask = new Task(userInput);
+                tasks.add(newTask);
+                System.out.println("____________________________________________________________");
+                System.out.println(" added: " + userInput);
```
ravi-viswa105: I think the indentation should be adjusted here

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+        return "[" + getStatusIcon() + "] " + description;
+    }
+}
+
+public class Astra {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> tasks = new ArrayList<>();
+
+        System.out.println("____________________________________________________________");
```
ravi-viswa105: I think a few print statements are not indented according to the standard

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+        System.out.println(" Hello! I'm Astra!");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        while (true) {
+            String userInput = scanner.nextLine().trim(); // Read and trim input
```
ravi-viswa105: I think that the comment is not required as the code is already clear

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+class Task {
```
rodi-314: I feel that the classes can be placed in separate files for clarity :))
```suggestion
class Task {
```

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // "X" for done, " " for not done
```
rodi-314: I like the clear inline comment :)
```suggestion
        return (isDone ? "X" : " "); // "X" for done, " " for not done
```

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> tasks = new ArrayList<>();
+
+        System.out.println("____________________________________________________________");
+        System.out.println(" Hello! I'm Astra!");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        while (true) {
```
rodi-314: I think a separate boolean can be created and passed into the while loop; using while (true) is not good practice in general as it can cause infinite loops! :0

src/main/java/Astra.java
```diff
@@ -0,0 +1,101 @@
...
+                    System.out.println(" No tasks added yet.");
+                } else {
+                    System.out.println(" Here are the tasks in your list:");
+                    for (int i = 0; i < tasks.size(); i++) {
+                        System.out.println(" " + (i + 1) + ". " + tasks.get(i));
+                    }
+                }
+                System.out.println("____________________________________________________________");
+            } else if (userInput.startsWith("mark ")) {
+                try {
```
rodi-314: I like the use of try-catch statements to detect invalid user inputs!
```suggestion
                try {
```

rodi-314 (commented):
Hi steve LOL good job overall just some minor comments to review :))

src/main/java/Piggy.java
```diff
@@ -0,0 +1,78 @@
...
+        System.out.println("    '|  .'     '.  |'");
+        System.out.println("     \\  | () () |  /");
+        System.out.println("    / '..     .'.' \\");
+        System.out.println("   /' / \\\"\"\"/ \\ \\");
+        System.out.println("  (/ /    \"\"\"    \\ \\)");
+        System.out.println("   ^               ^");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        while (true) {
```
bljhty: you may like to use a clearer exit condition ie a boolean variable instead. while(true) is generally not advisable

src/main/java/Piggy.java
```diff
@@ -0,0 +1,78 @@
...
+        System.out.println("(//| /             \\ |\\\\)");
+        System.out.println(" ^ \\ |    _..._    | / ^");
+        System.out.println("    '|  .'     '.  |'");
+        System.out.println("     \\  | () () |  /");
+        System.out.println("    / '..     .'.' \\");
+        System.out.println("   /' / \\\"\"\"/ \\ \\");
+        System.out.println("  (/ /    \"\"\"    \\ \\)");
+        System.out.println("   ^               ^");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
```
bljhty: In general, all these are considered as magic strings which are not encouraged, you may consider making them a constant instead?

src/main/java/Piggy.java
```diff
@@ -0,0 +1,78 @@
...
+
+        while (true) {
+            String userInput = scanner.nextLine().trim();
+
+            if (userInput.equals("bye")) {
+                System.out.println("____________________________________________________________");
+                System.out.println(" Bye. Hope to see you again soon!");
+                System.out.println("____________________________________________________________");
```
bljhty: similarly, magic strings should be made into constants as much as possible

src/main/java/Piggy.java
```diff
@@ -0,0 +1,78 @@
...
+                break;
+            } else if (userInput.equals("list")) {
+                taskList.listTasks();
+            } else if (userInput.startsWith("mark ")) {
+                try {
+                    int index = Integer.parseInt(userInput.substring(5)) - 1;
```
bljhty: the use of magic numbers are also not encouraged

src/main/java/Piggy.java
```diff
@@ -0,0 +1,78 @@
...
+                    taskList.markTask(index, false);
+                } catch (NumberFormatException e) {
+                    System.out.println(" Please enter a valid number.");
+                }
+            } else if (userInput.startsWith("todo ")) {
+                taskList.addTask(new ToDo(userInput.substring(5)));
+            } else if (userInput.startsWith("deadline ")) {
+                String[] parts = userInput.substring(9).split(" /by ", 2);
+                if (parts.length < 2) {
+                    System.out.println(" Invalid format! Use: deadline <description> /by <time>");
```
bljhty: magic strings

src/main/java/TaskList.java
```diff
@@ -0,0 +1,50 @@
...
+        this.tasks = new ArrayList<>();
+    }
+
+    public void addTask(Task task) {
+        tasks.add(task);
+        System.out.println("____________________________________________________________");
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("   " + task);
+        System.out.println(" Now you have " + tasks.size() + " tasks in the list.");
+        System.out.println("____________________________________________________________");
```
bljhty: magic strings can be made into constants

PR #174, opened 2025-02-05
[yangyi-zhu] iP

yangyi-zhu:


src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class Main {
+    public static final String DIVIDER = "--------------------------";
+    public static final String NL = System.lineSeparator(); // New Line
+
+    private static int count = 0;
+    private static final Task[] list = new Task[100];
+
+    /**
+    public static void echo(String msg) {
```
LEESY02: The commented code seems to be a snippet of code from a previous Level, perhaps it could be removed as it does not help the readers understand what is going on with your code.

src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
...
+                "-------------------------- \n");
+        Scanner newMsg = new Scanner(System.in);
+        echo(newMsg.nextLine());
+    }
+     */
+
+    public static void checkInput(String msg) {
+        if (msg.equals("list")) {
+            System.out.println(NL + DIVIDER + NL + "Tasks:");
+            for (int i = 0; i < count; i++) {
```
LEESY02: Maybe you could refactor each of the if cases into a method

Instead of using multiple lines of else if, perhaps having a single case() switch will be easier to read as well

LEESY02: e.g.: the lines of code for the list function could be
if (msg.equals("list")) {
    printList();
}
To simplify the method in general and making it easier to read

src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
...
+                "You now have " + (count + 1) + " tasks in the list." +
+                NL + DIVIDER + NL
+            );
+            count++;
+        }
+    }
+
+    public static void main(String[] args) {
+        Scanner msg = new Scanner(System.in);
+        System.out.println(
```
LEESY02: I think it would be easier to read if each of these print messages are encapsulated in a method, to make the main method shorter and more concise

LEESY02: for e.g.:  
        Scanner msg = new Scanner(System.in);
        printIntroMessage();
        .......
        printOutroMessage();
}

private static void printIntroMessage() {
        sout("This is an intro message :D");
}
private static void printOutroMessage() {
        sout("This is an outro message :D");
}

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
+public class Task {
+    private String desc;
+    private boolean isDone;
+
+    // Type 0: To-do (default)
```
LEESY02: Perhaps instead of using int type to identify the different sub tasks, could have Task as a parent class and the other subtasks as a children class

i.e.:
public class Task {
}
public class ToDo extends Task {
}

Using super classes' toString() in the child's classes' toString is an idea to automate the printing of each task
Task's toString: "[ ] Task A"
Child's toString: "[T]" + super.toString()
Of course this is a very simplified example, it is up to you on how you implement it

src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
+import java.util.Scanner;
```
LEESY02: Overall it was a good read with all the functionalities, there is only a few things to touch up :D

src/main/java/Baguette.java
```diff
@@ -1,4 +1,4 @@
-public class Duke {
+public class Baguette {
```
LEESY02: Nice class name, oui oui baguette

LEESY02 (commented):
I think majority of the code looks good! I gave some minor edit suggestions, to give them some consideration :D

src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final Task[] list = new Task[100];
+
+    /**
+    public static void echo(String msg) {
+        if (msg.equals("bye")) return;
+        System.out.println("\n" + "-------------------------- \n" +
+                msg + "\n" +
+                "-------------------------- \n");
+        Scanner newMsg = new Scanner(System.in);
+        echo(newMsg.nextLine());
```
Rishi7830: This commented code looks like a code of the previous level so remove it :) for better code quality

src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
...
+                NL + DIVIDER + NL +
+                "Got it! The following task has been marked as incomplete:" +
+                NL + "  " + list[index - 1] + NL +
+                DIVIDER + NL
+            );
+        } else {
+            int type = 0; // defaults to to-do if no prefix
+            int index = 0;
+
+            if (msg.startsWith("todo ")) {
```
Rishi7830: Good job on not using the arrowhead style code style

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
...
+    // Type 1: Deadline
+    // Type 2: Event
+    private int type;
+
+    public Task(String description, int type) {
```
Rishi7830: Good job on structuring the code logically

src/main/java/Main.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+
+    public static void checkInput(String msg) {
+        if (msg.equals("list")) {
+            System.out.println(NL + DIVIDER + NL + "Tasks:");
+            for (int i = 0; i < count; i++) {
+                System.out.println((i + 1) + ". " + list[i]);
+            }
+            System.out.println(DIVIDER + NL);
+        } else if (msg.startsWith("mark ")) {
```
Rishi7830: Good job on not using the arrowhead style code

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
...
+        this.isDone = true;
+    }
+
+    public void unmark() {
+        this.isDone = false;
+    }
+
+    @Override
+    public String toString() {
+        String typeBox = new String("");
```
Rishi7830: I like your code structure

Rishi7830 (commented):
Good Job

src/main/java/Main.java
```diff
@@ -0,0 +1,110 @@
+import java.util.Scanner;
+
+public class Main {
+    public static final String DIVIDER = "--------------------------";
+    public static final String NL = System.lineSeparator(); // New Line
+
+    private static int count = 0;
+    private static final Task[] list = new Task[100];
+
+    public static void checkInput(String msg) {
```
aditichadha1310: Avoid lengthy methods. Keep the length of each method to about 30 lines of code.

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public void unmark() {
+        this.isDone = false;
+    }
+
+    public abstract String getType();
+
+    @Override
+    public String toString() {
+        return getType() + (isDone ? "[✓] " : "[ ] ") + desc;
```
aditichadha1310: Avoid usage of magic strings.

src/main/java/Main.java
```diff
@@ -0,0 +1,110 @@
...
+                            NL + DIVIDER + NL +
+                                    "The supported types and corresponding formats are as follows:" + NL +
+                                    "To-Dos: todo [description]" + NL +
+                                    "Deadlines: deadline [description] ddl: [DD-MM-YY HH:MM]" + NL +
+                                    "Events: event [description] from: [DD-MM-YY HH:MM] to: [DD-MM-YY HH:MM]" + NL +
+                                    NL + "Examples:" + NL +
+                                    "todo Do the laundry" + NL +
+                                    "deadline CS2113 Increment L5 ddl: 14-02-25 16:00" + NL +
+                                    "event Josh's birthday party from: 03-03-25 15:00 to: 04-03-25 00:00" +
+                                    NL + DIVIDER + NL
```
aditichadha1310: Since the print statements are common for many commands -  you can consider creating a function instead of repeating the same code lines for each command

PR #175, opened 2025-02-06
[JonathanOngJoe] iP

Jonathan2745:


src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public void markAsNotDone() {
+        isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "[X] " : "[ ] ") + description;
+    }
+}
```
Li-JunXian: Well coded! Simple and straight forward, maybe perhaps we can try creating a separate class file for "Task" so everything is not placed in one single java file.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
...
+
+class Bob {
+    private final Scanner scanner = new Scanner(System.in);
+    private final String name = "Bob";
+    private final Task[] tasks = new Task[100];
+    private int taskCount = 0;
+
+    public void start() {
+        System.out.println("____________________________________________________________");
```
Li-JunXian: Perhaps the horizontal lines can be assigned to a variable for easier reference and usage in the future as compared to inputting it every time.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
...
+                if (taskCount == 0) {
+                    System.out.println(" No tasks added yet.");
+                } else {
+                    for (int i = 0; i < taskCount; i++) {
+                        System.out.println((i + 1) + ". " + tasks[i].getStatusIcon());
+                    }
+                }
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("mark ")) {
+                int index = Integer.parseInt(input.substring(5)) - 1;
```
Li-JunXian: I think it will be much more clearer to include comments on the meaning of "input.substring(5)) - 1" statement so that when looking back on the code in the future, everyone will understand the significance of the "- 1" for instance.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
...
+                    System.out.println("   " + tasks[index].getStatusIcon());
+                    System.out.println("____________________________________________________________");
+                }
+            } else if (input.startsWith("unmark ")) {
+                int index = Integer.parseInt(input.substring(7)) - 1;
+                if (index >= 0 && index < taskCount) {
+                    tasks[index].markAsNotDone();
+                    System.out.println("____________________________________________________________");
+                    System.out.println(" OK, I've marked this task as not done yet:");
+                    System.out.println("   " + tasks[index].getStatusIcon());
```
Li-JunXian: I think it is best not to start a System.out.printlin with a space or a tab, perhaps you can place the space and tab at the previous print statement's tail.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
...
+                    System.out.println(" No tasks added yet.");
+                } else {
+                    for (int i = 0; i < taskCount; i++) {
+                        System.out.println((i + 1) + ". " + tasks[i].getStatusIcon());
+                    }
+                }
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("mark ")) {
+                int index = Integer.parseInt(input.substring(5)) - 1;
+                if (index >= 0 && index < taskCount) {
```
Li-JunXian: I think this line appears quite often, perhaps we could position this line with the higher-up if statement so that it can be checked earlier for the program to determine if it is a valid input or not and then execute the correct corresponding actions without doing any redundant work.

Li-JunXian (commented):
Great work overall, your code has a very clear logic chain. It feels very smooth reading it.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
+import java.util.Scanner;
+
+class Task {
```
Deanson-Choo: Maybe you can consider defining your classes such as Task in a separate file.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,97 @@
...
+                if (taskCount < 100) {
+                    tasks[taskCount++] = new Task(input);
+                    System.out.println("____________________________________________________________");
+                    System.out.println(" added: " + input);
+                    System.out.println("____________________________________________________________");
+                } else {
+                    System.out.println(" Task list is full!");
+                }
+            }
+        }
```
Deanson-Choo: Perhaps you can consider adding another class to manage your tasks to neaten the code

src/main/java/BobChat.java
Deanson-Choo: Perhaps you can consider combining markAsNotDone and markAsDone under a single function such as setStatus. Also, should there be a getter function for description and isDone?

src/main/java/BobChat.java
Deanson-Choo: Perhaps considering assigning the horizontal line to a variable for easier usage

src/main/java/BobChat.java
Deanson-Choo: Perhaps you can consider adding comments to each of your conditions too!

Deanson-Choo (commented):
Good Job Overall! A few touchups will do!

src/main/java/BobChat.java
```diff
@@ -0,0 +1,135 @@
...
+        while (true) {
+            String input = scanner.nextLine().trim();
+            String[] inputParts = input.split(" ", 2);
+            String command = inputParts[0].toLowerCase();
+            String arguments = (inputParts.length > 1) ? inputParts[1] : "";
+
+            try {
+                switch (command) {
+                    case "bye":
+                        System.out.println("____________________________________________________________");
```
aditichadha1310: Since this line of code is repeating at multiple instances, you might want to save it in a variable or make it a function which you can reference.

src/main/java/BobChat.java
```diff
@@ -0,0 +1,135 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+import Errors.*;
+
+class Bob {
+    private final Scanner scanner = new Scanner(System.in);
+    private final ArrayList<Task> tasks = new ArrayList<>();
+
+    public void start() throws InputExceptions {
```
aditichadha1310: Avoid lengthy methods. try to keep each method to about 30 lines of code.

aditichadha1310 (commented):
overall, good work!

PR #180, opened 2025-02-06
[Liu Yishan] iP

lys2333:


src/main/java/Lys.java
```diff
@@ -1,3 +1,81 @@
+import java.util.Scanner;
+
+abstract class Task {
```
Isaaclks7, line 8: Correct naming convention of your class in PascalCase.

src/main/java/Lys.java
```diff
@@ -1,3 +1,81 @@
...
+        this.isDone = false;
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void unmarkAsDone() {
+        this.isDone = false;
+    }
```
Isaaclks7: Good use of camelCase for methods

src/main/java/Lys.java
```diff
@@ -6,12 +84,79 @@ public static void main(String[] args) {
...
+            } else if (words[0].equalsIgnoreCase("event")) {
+                String[] parts = words[1].split(" /from | /to ");
+                tasks[taskCount] = new Event(parts[0], parts[1], parts[2]);
+                taskCount++;
+                System.out.println("____________________________________________________________");
+                System.out.println("Got it. I've added this task:");
+                System.out.println("  " + tasks[taskCount - 1]);
+                System.out.println("Now you have " + taskCount + " tasks in the list.");
+                System.out.println("____________________________________________________________");
+            }
```
Isaaclks7: Could consider using switch-case instead of multiple if-else statements

src/main/java/Lys.java
```diff
@@ -6,12 +84,79 @@ public static void main(String[] args) {
...
+            if (input.equalsIgnoreCase("bye")) {
+                System.out.println("____________________________________________________________");
+                System.out.println("Bye. Hope to see you again soon!");
+                System.out.println("____________________________________________________________");
+                break;
+            } else if (input.equalsIgnoreCase("list")) {
+                System.out.println("____________________________________________________________");
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println((i + 1) + "." + tasks[i]);
```
Isaaclks7: Could rename i to taskIndex for readability. Additionally, you can start from i = 1 instead of i = 0.

Isaaclks7 (commented):
Overall good naming conventions and relevant spacings. However, you may want to place your different classes into separate files for better organisation.

src/main/java/Lys.java
```diff
@@ -0,0 +1,188 @@
...
+            String command = words[0].toLowerCase();
+
+            try {
+                switch (command) {
+                    case "bye":
+                        System.out.println("____________________________________________________________");
+                        System.out.println("Bye. Hope to see you again soon!");
+                        System.out.println("____________________________________________________________");
+                        scanner.close();
+                        return;
```
xenthm: Do try to use the formatting for `switch` statements outlined in [the Java coding standards for this course](https://se-education.org/guides/conventions/java/index.html#layout).

src/main/java/Lys.java
```diff
@@ -0,0 +1,188 @@
...
+                            System.out.println("____________________________________________________________");
+                            System.out.println("Nice! I've marked this task as done:");
+                        } else {
+                            tasks[index].unmarkAsDone();
+                            System.out.println("____________________________________________________________");
+                            System.out.println("OK, I've marked this task as not done yet:");
+                        }
+                        System.out.println("  " + tasks[index]);
+                        System.out.println("____________________________________________________________");
+                        break;
```
xenthm: Good job realising that the `mark` and `unmark` commands are very similar. Consider extracting the logic out to another method so that you can avoid placing the `if`-`else` block within the `case` block for `mark` and `unmark`, which basically does the same thing.

src/main/java/Lys.java
```diff
@@ -0,0 +1,188 @@
...
+                        if (taskCount == 0) {
+                            System.out.println("Your task list is empty.");
+                        } else {
+                            System.out.println("Here are the tasks in your list:");
+                            for (int i = 0; i < taskCount; i++) {
+                                System.out.println((i + 1) + "." + tasks[i]);
+                            }
+                        }
+                        System.out.println("____________________________________________________________");
+                        break;
```
xenthm: Perhaps you can extract some of the logic in your cases out into methods in order to [reduce deep nesting](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting) and prevent "arrowhead code" from developing.

PR #181, opened 2025-02-06
[crl006] iP

CRL006:


src/main/java/Bobby.java
```diff
@@ -3,40 +3,83 @@ public class Bobby {
...
         int markIndex;
+        int stringLength;
         String greeting = "Hello! I'm Bobby! \nWhat can I do for you? \n";
         System.out.println(greeting);
         String[] tasks = new String[100];
         Boolean[] isDone = new Boolean[100];
+        int[] taskType = new int[100];
         Scanner scanner = new Scanner(System.in);  // Create a Scanner object
         String userInput = scanner.nextLine();
         while (!userInput.equals("bye")) {
```
kemilii: I like your way of using "if equals 'bye'" directly as the while condition

src/main/java/Bobby.java
```diff
@@ -3,40 +3,83 @@ public class Bobby {
...
-                    System.out.print((i+1) + ". ");
+                    System.out.print((i + 1) + ". ");
+                    if (taskType[i] == 0) {
+                        System.out.print("[T]");
+                    } else if (taskType[i] == 1){
+                        System.out.print("[D]");
+                    } else {
+                        System.out.print("[E]");
+                    }
                     if (isDone[i]) {
```
kemilii: Any reason you did it using "if-else" instead of "switch"? Because personally if the condition is the same single "variable", I feel "switch" would be more clear

src/main/java/Bobby.java
```diff
@@ -3,40 +3,83 @@ public class Bobby {
...
                         System.out.print("[X] ");
-                    }
-                    else {
+                    } else {
                         System.out.print("[ ] ");
                     }
                     System.out.println(tasks[i]);
                 }
+            } else if (userInput.contains("todo")){
+                stringLength = userInput.length();
```
kemilii: I like the way of assigning the value of length to a variable, making it more clear. But perhaps a more intuitive variable name here?

src/main/java/Bobby.java
```diff
@@ -3,40 +3,83 @@ public class Bobby {
...
+                System.out.println("Got it. I've added this task: \n  [T][ ] " + userInput.substring(5, stringLength) + "\nNow you have " + (tasksCounter + 1) + " tasks in the list.");
+                tasks[tasksCounter] = userInput.substring(5, stringLength);
+                isDone[tasksCounter] = false;
+                taskType[tasksCounter] = 0;
+                tasksCounter++;
+            } else if (userInput.contains("deadline")){
+                stringLength = userInput.length();
+                int indexOfSlash = userInput.indexOf("/");
```
kemilii: Not sure if using a simple "/" will misuse things other than "/by" to split the string

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+public class Bobby {
+    public static void main(String[] args) {
+        int tasksCounter = 0;
+        int markIndex;
+        int stringLength;
+        String greeting = "Hello! I'm Bobby! \nWhat can I do for you? \n";
+        System.out.println(greeting);
+        String[] tasks = new String[100];
+        Boolean[] isDone = new Boolean[100];
+        int[] taskType = new int[100];
```
lwenyi1: Create a separate `Task` class for the tasks

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+                        System.out.print("[E]");
+                    }
+                    if (isDone[i]) {
+                        System.out.print("[X] ");
+                    } else {
+                        System.out.print("[ ] ");
+                    }
+                    System.out.println(tasks[i]);
+                }
+            } else if (userInput.contains("todo")){
```
lwenyi1: Consider creating a separate `Todo` class that extends from a `Task` class, likewise for deadline and event

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
+import java.util.Scanner;  // Import the Scanner class
+public class Bobby {
+    public static void main(String[] args) {
+        int tasksCounter = 0;
```
lwenyi1: Consider changing the name to `taskCounter`, more concise and conventional

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+        Boolean[] isDone = new Boolean[100];
+        int[] taskType = new int[100];
+        Scanner scanner = new Scanner(System.in);  // Create a Scanner object
+        String userInput = scanner.nextLine();
+        while (!userInput.equals("bye")) {
+            if (userInput.equals("list")) {
+                System.out.println("Here are the tasks in your list: ");
+                for (int i = 0; i < tasksCounter; i++) {
+                    System.out.print((i + 1) + ". ");
+                    if (taskType[i] == 0) {
```
lwenyi1: Unnecessarily complicated to check for task type like that, use an attribute `taskType` in a `Task` class to store instead

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+                tasks[tasksCounter] = userInput.substring(9, indexOfSlash-1) + " (by: " + userInput.substring(indexOfSlash+4, stringLength) + ")";
+                isDone[tasksCounter] = false;
+                taskType[tasksCounter] = 1;
+                System.out.println("Got it. I've added this task: \n  [D][ ] " + tasks[tasksCounter] + "\nNow you have " + (tasksCounter + 1) + " tasks in the list.");
+                tasksCounter++;
+            } else if (userInput.contains("event")){
+                stringLength = userInput.length();
+                int indexOfFirstSlash = userInput.indexOf("/");
+                int indexOfLastSlash = userInput.lastIndexOf("/");
+                tasks[tasksCounter] = userInput.substring(6, indexOfFirstSlash-1) + " (from: " + userInput.substring(indexOfFirstSlash+6, indexOfLastSlash-1) + " to: " + userInput.substring(indexOfLastSlash + 4, stringLength) + ")";
```
lwenyi1: Explore using the `.split` function instead to make code less complicated

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
+import java.util.Scanner;  // Import the Scanner class
+public class Bobby {
+    public static void main(String[] args) {
+        int tasksCounter = 0;
+        int markIndex;
```
lwenyi1: not immediately clear what `markIndex` does, consider renaming to `markTaskIndex` or documenting

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+        int stringLength;
+        String greeting = "Hello! I'm Bobby! \nWhat can I do for you? \n";
+        System.out.println(greeting);
+        String[] tasks = new String[100];
+        Boolean[] isDone = new Boolean[100];
```
lwenyi1: `isDone` would be a clear name if it was part of a Task class, but inside main it is not descriptive enough

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+                } else if (taskType[markIndex] == 1){
+                    System.out.print("[D]");
+                } else {
+                    System.out.print("[E]");
+                }
+                System.out.println("[X] " + tasks[markIndex]);
+            }
+            userInput = scanner.nextLine();
+        }
+        String endChat = "Bye. Hope to see you again soon!";
```
lwenyi1: `endChat` not really needed considering you only use it once

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+                if (taskType[markIndex] == 0) {
+                    System.out.print("[T]");
+                } else if (taskType[markIndex] == 1){
+                    System.out.print("[D]");
+                } else {
+                    System.out.print("[E]");
+                }
+                System.out.println("[ ] " + tasks[markIndex]);
+            } else if (userInput.contains("mark")) {
+                String[] splitMark = userInput.split(" ");
```
lwenyi1: `splitMarker` would be clearer

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
+import java.util.Scanner;  // Import the Scanner class
+public class Bobby {
+    public static void main(String[] args) {
+        int tasksCounter = 0;
+        int markIndex;
+        int stringLength;
+        String greeting = "Hello! I'm Bobby! \nWhat can I do for you? \n";
```
lwenyi1: Not really needed IMO but if really must be included can be changed to `GREETING` since it's a constant string

lwenyi1 (changes requested):
Nice attempt, very creative approach to creating `Bobby`.

Highly recommend following the [OOP paradigm](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/oop.html) instead.

*Apologies, accidentally classified as `Request changes`, should be `Comment`

src/main/java/Bobby.java
```diff
@@ -3,40 +3,83 @@ public class Bobby {
...
+                markIndex = Integer.parseInt(splitMark[1]) - 1;
+                isDone[markIndex] = false;
+                System.out.println("OK, I've marked this task as not done yet:");
+                if (taskType[markIndex] == 0) {
+                    System.out.print("[T]");
+                } else if (taskType[markIndex] == 1){
+                    System.out.print("[D]");
+                } else {
+                    System.out.print("[E]");
+                }
```
kemilii: Very effective way to implement deadline, todo and event. But perhaps the usage of inheritance will make the code more readable and also easier to debug (modularity)?

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
+import java.util.Scanner;  // Import the Scanner class
+public class Bobby {
+    public static void main(String[] args) {
+        int tasksCounter = 0;
+        int markIndex;
+        int stringLength;
+        String greeting = "Hello! I'm Bobby! \nWhat can I do for you? \n";
+        System.out.println(greeting);
+        String[] tasks = new String[100];
```
Lyam-T: try to avoid the magic number

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+        Boolean[] isDone = new Boolean[100];
+        int[] taskType = new int[100];
+        Scanner scanner = new Scanner(System.in);  // Create a Scanner object
+        String userInput = scanner.nextLine();
+        while (!userInput.equals("bye")) {
+            if (userInput.equals("list")) {
```
Lyam-T: I don't like the arrowhead indentation, maybe can be simplified

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+                }
+                System.out.println("[ ] " + tasks[markIndex]);
+            } else if (userInput.contains("mark")) {
+                String[] splitMark = userInput.split(" ");
+                markIndex = Integer.parseInt(splitMark[1]) - 1;
+                isDone[markIndex] = true;
+                System.out.println("Nice! I've marked this task as done:");
+                if (taskType[markIndex] == 0) {
+                    System.out.print("[T]");
+                } else if (taskType[markIndex] == 1){
```
Lyam-T: same for this taskType, I don't like the magic number, perhaps can use a constant string to represent instead

src/main/java/Bobby.java
```diff
@@ -0,0 +1,89 @@
...
+                stringLength = userInput.length();
+                int indexOfFirstSlash = userInput.indexOf("/");
+                int indexOfLastSlash = userInput.lastIndexOf("/");
+                tasks[tasksCounter] = userInput.substring(6, indexOfFirstSlash-1) + " (from: " + userInput.substring(indexOfFirstSlash+6, indexOfLastSlash-1) + " to: " + userInput.substring(indexOfLastSlash + 4, stringLength) + ")";
+                System.out.println("Got it. I've added this task: \n  [E][ ] " + tasks[tasksCounter] + "\nNow you have " + (tasksCounter + 1) + " tasks in the list.");
+                isDone[tasksCounter] = false;
+                taskType[tasksCounter] = 2;
+                tasksCounter++;
+            } else if (userInput.contains("unmark")) {
+                String[] splitMark = userInput.split(" ");
```
Lyam-T: I dont like the name splitMark, maybe it can be more specific

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+    public Deadline(String item, String dueDate, boolean isDone) {
+        super(item, isDone);
+        this.item = item;
+        this.dueDate = dueDate;
+    }
+    @Override
+    public String toString() {
+        return "D | " + (isDone ? "done" : "not done") + " | " + item + " | by: " + dueDate;
+    }
+    @Override
```
Hackin7, line 49: Add an empty line between methods for better readability

PR #192, opened 2025-02-07
[paklongchiu] iP

paklongchiu:


src/main/java/Elyk.java
```diff
@@ -0,0 +1,79 @@
...
+    public static int taskCounter = 0;
+    public static int taskNum = 0;
+    public static String input = "";
+    public static Task[] taskList = new Task[100];
+    public static Scanner command = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        greet();
+
+        while (true) {
```
Ashertan256, line 38: Instead of while(true), maybe you can consider using this instead:

do {
  // code to run
}
while (!input.equal("bye"));

src/main/java/Elyk.java
```diff
@@ -0,0 +1,79 @@
...
+    public static void markTaskDone(int taskNum) {
+        taskList[taskNum - 1].markAsDone();
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println("  [X] " + taskList[taskNum - 1].description);
+    }
+
+    public static void markTaskNotDone(int taskNum) {
+        taskList[taskNum - 1].markAsNotDone();
+        System.out.println("OK, I've marked this task as not done yet:");
+        System.out.println("  [ ] " + taskList[taskNum - 1].description);
```
Ashertan256: Would using taskList[i].getStatusIcon (as you used below) be more consistent?

src/main/java/Elyk.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public static void sayBye() {
+        String bye = "Bye. Hope to see you again soon!";
+        System.out.println(bye);
+    }
+
+    public static void updateInput() {
+        input = command.nextLine();
+        if (input.matches(".*\\d$")) {
```
Ashertan256: A comment could be added to describe the regex pattern to be clearer

src/main/java/Elyk.java
```diff
@@ -0,0 +1,79 @@
...
+            String[] words = input.split(" ");
+            input = words[0];
+            taskNum = Integer.parseInt(words[1]);
```
Ashertan256: This could potentially have issues if the second word is not a number. Attempting to parseInt a String may crash the program/cause an exception

src/main/java/Elyk.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public static void markTaskDone(int taskNum) {
+        taskList[taskNum - 1].markAsDone();
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println("  [X] " + taskList[taskNum - 1].description);
+    }
+
+    public static void markTaskNotDone(int taskNum) {
+        taskList[taskNum - 1].markAsNotDone();
```
Ashertan256: This seems like a redundant method, is it possible to call a single method without chaining multiple methods that do the same thing?

Ashertan256 (commented):
Overall, the structure and flow of the code seems a little redundant at certain parts. But the format looks to be consistent with the course!

src/main/java/Elyk.java
```diff
@@ -0,0 +1,110 @@
...
+            int byPos = input.indexOf("/by");
+            by = input.substring(byPos + 4);
+            description = input.substring(9, byPos - 1);
+            input = "deadline";
+        } else if (input.startsWith("event")) {
+            int fromPos = input.indexOf("/from");
+            int toPos = input.indexOf("/to");
+            to = input.substring(toPos + 4);
+            from = input.substring(fromPos + 6, toPos - 1);
+            description = input.substring(6, fromPos - 1);
```
thomasjlalba: Maybe you can[ refactor the magic literals](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-magic-numbers) throughout your code to make it easier to understand

src/main/java/Elyk.java
```diff
@@ -0,0 +1,110 @@
...
+            input = "event";
+        }
```
thomasjlalba: Maybe you can [include the final else statement](https://nus-cs2113-ay2425s2.github.io/website/se-book-adapted/chapters/codeQuality.html#use-the-default-branch) for error detection

src/main/java/Elyk.java
```diff
@@ -0,0 +1,110 @@
...
+    public static void inputTask() {
+        System.out.println("Got it. I've added this task:");
+        System.out.println("  " + taskList[taskCounter - 1]);
+        System.out.println("Now you have " + taskCounter + " tasks in the list.");
+    }
+
+    public static void printTask() {
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < taskCounter; i++) {
+            System.out.println((i+1) + "." + taskList[i]);
```
thomasjlalba: Maybe you can standardise the use of spacing within your `i+1` to make it more consistent (`i + 1` instead)

src/main/java/Elyk.java
```diff
@@ -0,0 +1,110 @@
...
+                    printTask();
+                    break;
+                case "mark":
+                    markTaskDone(taskNum);
+                    break;
+                case "unmark":
+                    markTaskNotDone(taskNum);
+                    break;
+                case "todo":
+                    taskList[taskCounter++] = new Todo(description);
```
thomasjlalba: It might be more readable by separating the increment to the next line instead of combining it in the same line

src/main/java/Elyk.java
```diff
@@ -0,0 +1,110 @@
+import java.util.Scanner;
+
+public class Elyk {
+    public static int taskCounter = 0;
+    public static int taskNum = 0;
```
thomasjlalba: The naming of these variables might suggest that they are the same. Maybe you can consider renaming them to make them clearer and more distinct.

src/main/java/Elyk.java
```diff
@@ -0,0 +1,110 @@
...
+    public static String input = "";
+    public static String description = "";
+    public static String from = "";
+    public static String to = "";
+    public static String by = "";
+    public static Task[] taskList = new Task[100];
+    public static Scanner command = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        greet();
```
thomasjlalba: Good job on the abstraction of code to make it neater!

thomasjlalba (commented):
Good job thus far! Just some trivial suggestions to make it more readable and understandable for others.

PR #206, opened 2025-02-13
[rozaliesmit] iP

rozaliesmit:


src/main/java/Daisy.java
```diff
@@ -0,0 +1,269 @@
...
+    private static void invalidMarkTask(String input) {
+        try {
+            int taskNumber = Integer.parseInt(input.split(" ")[1]);
+            markTask(taskNumber);
+        } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
+            System.out.println("Oh no! Seems like the command format was invalid. Please use the format 'mark <task_number>' instead.");
+        }
+    }
+
+    private static void invalidUnmarkTask(String input) {
```
aditichadha1310: You might want to structure the classes into different folders.

src/main/java/Daisy.java
```diff
@@ -0,0 +1,269 @@
...
+            } else if (input.startsWith("unmark")) {
+                invalidUnmarkTask(input);
+            } else if (input.startsWith("todo")) {
+                invalidTodoTask(input);
+            } else if (input.startsWith("deadline")) {
+                invalidDeadlineTask(input);
+            } else if (input.startsWith("event")) {
+                invalidEventTask(input);
+            } else {
+                System.out.println("____________________________________________________________");
```
aditichadha1310: Avoid the usage of magic strings.

src/main/java/Daisy.java
```diff
@@ -0,0 +1,269 @@
...
+        System.out.println("____________________________________________________________");
+        System.out.println(" Nice! I've marked this task as done:");
+        System.out.println("   " + task);
+        System.out.println("____________________________________________________________");
+    }
+
+    // Unmark task
+    private static void unmarkTask(int taskNumber) {
+        if (taskNumber > 0 && taskNumber <= taskCount) {
+            tasks[taskNumber - 1].markAsNotDone();
```
aditichadha1310: Avoid the usage of magic numbers.

PR #208, opened 2025-02-14
[kagiura] iP

kagiura:


src/main/java/Wen.java
```diff
@@ -1,19 +1,34 @@
 import java.util.Scanner;
 
 public class Wen {
+
+    private static int taskCount = 0;
+    private static Task[] tasks = new Task[100];
```
Alfred-Goh02: avoid using magic numbers, declare 100 as a constant like TASKS_SIZE

src/main/java/Wen.java
```diff
@@ -1,19 +1,34 @@
...
+                case "list":
+                    printTasks();
+                    break;
+
+
+
                 default:
-                    System.out.println(command);
+                    addTask(input);
+//                    System.out.println("Unknown command \""+command+"\"! Double check your message and try running again~");
```
Alfred-Goh02: Delete the codes that are commented and not used.

src/main/java/Wen.java
```diff
@@ -1,19 +1,34 @@
...
         Scanner in = new Scanner(System.in);
 
-        while (!command.equals("bye")) {
-            command = in.nextLine();
+        while (!input.equals("bye")) {
+            input = in.nextLine();
+            final String command = input.split(" ", 2)[0];
+            final String commandArgs = input.split(" ", 2).length > 1 ? input.split(" ", 2)[1] : "";
+
             switch (command) {
```
Alfred-Goh02: There should be no indentation for case clauses, refer to the Java Coding standard.

src/main/java/Wen.java
```diff
@@ -1,19 +1,34 @@
...
     public static void main(String[] args) {
         initializeAndGreet();
 
-        String command = "";
+        String input = "";
         Scanner in = new Scanner(System.in);
 
-        while (!command.equals("bye")) {
-            command = in.nextLine();
+        while (!input.equals("bye")) {
```
Alfred-Goh02: The nested loop is not correctly indented, and consider whether you need the outer loop or not.

